PR #35, opened 2022-06-07
fix: fixes improper param type, closes #34

### LINKED ISSUE #34: [bug] bulkCancelOrders is broken
calling 

```
seaport.bulkCancelOrders()
```

results in error

Screen Shot 2022-06-07 at 08 46 14

because of improper type conversion

[here](https://github.com/ProjectOpenSea/seaport-js/blob/1f285ec3594446f5fb1754bed0bd056037b812fd/src/seaport.ts#L401 ) 
an empty array is passed 

[here](https://github.com/ProjectOpenSea/seaport-js/blob/1f285ec3594446f5fb1754bed0bd056037b812fd/src/utils/usecase.ts#L60) last item of this empty array is taken which is undefined

and [here](https://github.com/ProjectOpenSea/seaport-js/blob/1f285ec3594446f5fb1754bed0bd056037b812fd/src/utils/usecase.ts#L49) this undefined is passed to Object.keys which can accept an only object.

I can create PR with a fix. just tell me if defaulting `obj` to `{}` in `instanceOfOverrides` would be appropriate

TrejGun:


sunwrobert:
don't think this is an issue anymore, we can close

PR #29, opened 2022-05-27
chore: update eslint plugins

TrejGun:
fixes warning on installation
same as this one 
https://github.com/ProjectOpenSea/seaport/pull/288

PR #39, opened 2022-06-12
chore: update eslint plugins

TrejGun:


TrejGun:
@sunwrobert something went wrong with #29 so i created new one

sunwrobert:
@TrejGun I think you need to rename the eslint rule to `n/no-unsupported-features/es-syntax`?

TrejGun:
yup, sorry for this

PR #56, opened 2022-06-24
Set opensea conduit as default conduit

arr00:


sunwrobert (commented):
I don't know if we want opensea's conduit key to be the default. We could add it to the known conduit keys mapping though, as well as expose the variables so others could use.

If you want to use this library for opensea-specific stuff, you're better off using `opensea-js`

arr00:
Yeah you are right

sunwrobert (approved):

PR #57, opened 2022-06-26
chore: add basic examples in the README. 

### LINKED ISSUE #51: Example in the Readme
Is it correct that opensea.js will be deprecated in the future in favor of this one? If so it would be great to have examples like opensea.js has of things like fulfilling sell orders. Also was there a decision to move from web3 to ethers, or will this also support web3?

sunwrobert:
Closes #51

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,73 @@ Look at the relevant definitions in `seaport.ts` in order to see the different f
 
 Many of the main core flows return _use cases_. What this means is that if you were to create an order (a la `createOrder`), the library helps perform the necessary balance and approval checks based on the `offer` of the order being created. If the `offerer` requires approvals on one asset contract, the `actions` field of the use case would contain an approval action that the user should execute first in order for the trade to succeed in the future.
 
+### Examples
+
+#### Listing an ERC-721 for 10 ETH and fulfilling it
+
+```
+const [offerer, fulfiller] = await ethers.getSigners();
+const { executeAllActions } = await seaport.createOrder({
+  offer: [
+    {
+      itemType: ItemType.ERC721,
+      token: "0x8a90cab2b38dba80c64b7734e58ee1db38b8992e",
+      identifier: "1",
+    }
+  ],
+  consideration: [
+    {
+      amount: parseEther("10").toString(),
+      recipient: offerer.address,
+    }
+  ],
+  accountAddress: offerer.address
+});
+
+const order = await executeAllActions();
+
+const { executeAllActions: executeAllFulfillActions } = await seaport.fulfillOrder({
+  order,
+  accountAddress: fulfiller.address,
+});
+
+const transaction = executeAllFulfillActions()
+```
+
+#### Making an offer for an ERC-721 for 10 WETH and fulfilling it
+
+```
+const [offerer, fulfiller] = await ethers.getSigners();
+const { executeAllActions } = await seaport.createOrder({
+  offer: [
+
+    {
+      amount: parseEther("10").toString(),
+      // WETH
+      token: "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2"
+    },
+  ],
+  consideration: [
+    {
+      itemType: ItemType.ERC721,
+      token: "0x8a90cab2b38dba80c64b7734e58ee1db38b8992e",
+      identifier: "1",
+      recipient: offerer.address
+    },
+  ],
+  accountAddress: offerer.address
+});
+
+const order = await executeAllActions();
+
+const { executeAllActions: executeAllFulfillActions } = await seaport.fulfillOrder({
+  order,
+  accountAddress: fulfiller.address,
+});
+
+const transaction = executeAllFulfillActions()
+```
+
 ## Contributing
 
 See [the contributing guide](CONTRIBUTING.md) for detailed instructions on how to get started with this project.


Benjythebee:
Tested this,
breaks and throws an bignumber error for `totalItemsAmount(currencies);`
https://github.com/ProjectOpenSea/seaport-js/blob/3d789593e54594acfeef88e83bb0b3a4c1570af9/src/utils/order.ts#L135

```js
Error: invalid BigNumber value (argument="value", value=undefined, code=INVALID_ARGUMENT, version=bignumber/5.6.1)
```

I am using startAmount and endAmount instead of `amount`; and this is the parameters:
image

Following your example, I don't see much of a difference between my dutch auction integration and your fixed price example

sunwrobert:
> Tested this, breaks and throws an bignumber error for `totalItemsAmount(currencies);`
> 
> https://github.com/ProjectOpenSea/seaport-js/blob/3d789593e54594acfeef88e83bb0b3a4c1570af9/src/utils/order.ts#L135
> 
> ```js
> Error: invalid BigNumber value (argument="value", value=undefined, code=INVALID_ARGUMENT, version=bignumber/5.6.1)
> ```
> 
> I am using startAmount and endAmount instead of `amount`; and this is the parameters: image
> 
> Following your example, I don't see much of a difference between my dutch auction integration and your fixed price example

this is because the input actually takes it as `amount` instead of `startAmount`, but perhaps we should just change it to be `startAmount` in general.

Benjythebee:
Switched `startAmount` to `amount` launched the signature request and returned the parameters and a signature hash;
so worked all good;

Now this might be an unrelated issue, but not entirely sure:
When I used the received parameters and signature hash from above with the api route to create a listing `https://api.opensea.io/v2/orders/ethereum/seaport/listings`
I got the following error:
```
{
errors: [
  "You provided an invalid conduit key: 0x0000000000000000000000000000000000000000000000000000000000000000, please use OpenSea's conduit key: 0x0000007b02230091a7ed01230072f7006a004d60a8d4e71d599b8104250f0000 for order placement"
 ]
}
```

sunwrobert:
That's because you should use opensea-js directly when trying to post OpenSea orders on OpenSea. this is aimed to be a general purpose library for any marketplace using Seaport. opensea-js uses this one under the hood and provides the opensea conduit key

Benjythebee:
Yeah, thanks for bearing my questions;

Been trying to use seaport directly as Opensea-js has a few issues atm

pguardiario:
Thanks @sunwrobert,
That's very helpful but I think you're using a weird ethers version and I'm having trouble getting this working in my React app with latest ethers. Can you show how to do the signers part with web3 or latest ethers?

sunwrobert:
> Thanks @sunwrobert, That's very helpful but I think you're using a weird ethers version and I'm having trouble getting this working in my React app with latest ethers. Can you show how to do the signers part with web3 or latest ethers?

what part specifically are you having trouble with?

pguardiario:
I get this error: ethers.getSigners is not a function
Apparently the new way is:
```
const provider = new ethers.providers.Web3Provider(window.ethereum);
const signer = new provider.getSigner();
```
But I think it's returning something different now.

sunwrobert:
> I get this error: ethers.getSigners is not a function Apparently the new way is:
> 
> ```
> const provider = new ethers.providers.Web3Provider(window.ethereum);
> const signer = new provider.getSigner();
> ```
> 
> But I think it's returning something different now.

Ah yeah, getSigners is not the important part in the example. You just need to pass in any address that your provider has access to.

pguardiario:
Ok that makes sense.
For the 2nd example I'm getting: 
`Error: The fulfiller does not have the balances needed to fulfill.`
Even though I have enough WETH.
Also you give examples for creating buy / sell order but I was really looking for fulfill existing sell order with ETH (Buy now in other words)

sunwrobert:
> Ok that makes sense. For the 2nd example I'm getting: `Error: The fulfiller does not have the balances needed to fulfill.` Even though I have enough WETH. Also you give examples for creating buy / sell order but I was really looking for fulfill existing sell order with ETH (Buy now in other words)

The example has buy now, that's what the `fulfillOrder` is for.

pguardiario:
Oh I see, I thought that was a bid because it was WETH. So the error will disregard the amount I send and use the price it's listed as? And that's why I'm getting the error?
Can we also get an example of create buy order in that case?
Is it possible to do collection orders like on the website?

sunwrobert:
> I guess I'm mostly confused about this part:
> 
> #### Making an offer for an ERC-721 for 10 WETH and fulfilling it
> Why would I fulfill an order that I made? And how would I fulfill an order that someone else made? In other words how would I buy a token that is currently listed on opensea.

You wouldnt be fulfilling an order that you made.. You pass in an order that someone else has made.. 

The example shows two sides of the same coin

sunwrobert:
@pguardiario the created order is simply an object. If you're talking about POSTing an order to OpenSea, please refer to `opensea-js`

pguardiario:
So I think it would be helpful to show an example of how to fulfill an existing sell order on opensea using seaport.js.
Opensea-js does an excellent job of doing this plainly in the README, so hopefully something like that will be made available in the future.

pguardiario:
I think a lot of people would be interested in fulfilling orders without opensea-js because that involves lots of unnecessary lookups and an API key.

landakram:
I'm just a passer-by, but @pguardiario, perhaps the thing to keep in mind is that, when you use opensea-js, you are using OpenSea's hosted, off-chain order-book. The reason you need an OpenSea API key and API look-ups to fulfill an order using opensea-js is that you must fetch the offerer's order from OpenSea's hosted order-book so that it may be submitted as part of the fulfiller's order to wyvern/seaport, where both orders are validated and the swap is executed.

When you're using seaport directly, there is no order-book: _you_ need to maintain the order-book through some other means.  The example in this repo's README leaves fetching the existing offerer order from some order-book as an exercise to the reader. In fact, it's not strictly required to have an order-book. You could build a product on top of seaport that e.g. encodes the offerer's order in a URL that the offerer can send directly to potential fulfillers. Or you could rely on seaport's on-chain order-book functionality (using `seaport.validate`). But in general, you will likely need some sort of off-chain storage that can be used to track / maintain an order-book.

pguardiario:
@landakram,
Yes I don't want to do any lookups, I will already know the price.

When I go to the seaport etherscan page it doesn't show the transaction method names and it can't decode the input, I don't understand why:
https://etherscan.io/address/0x00000000006c3852cbef3e08e8df289169ede581

PR #63, opened 2022-07-04
Added support for custom signer.

CesarAHernandez:
Added support for custom signer with Ethers.Wallet.


src/seaport.ts
```diff
@@ -82,6 +83,7 @@ export class Seaport {
   /**
    * @param provider - The provider to use for web3-related calls
    * @param considerationConfig - A config to provide flexibility in the usage of Seaport
+   * @param signer - The signer used to sign your accounts
    */
   public constructor(
     provider: providers.JsonRpcProvider,
```
sunwrobert, line 89: do people need to provide a `provider` usually if they're going for the custom signer approach?

Wondering if it's better to just make this `providerOrSigner`, and then assign `this.provider` based on the signer's provider if the provided object is a signer otherwise, just the provider.

src/seaport.ts
```diff
@@ -349,15 +354,14 @@ export class Seaport {
     counter: number,
     accountAddress?: string
   ): Promise<string> {
-    const signer = this.provider.getSigner(accountAddress);
+    const signer = this.signer ?? this.provider.getSigner(accountAddress);
```
sunwrobert, line 357: can we make this into a method that's just `_getSigner(accountAddress?: str)` that will return `this.signer` if signer is defined, otherwise fallback to the provider's `getSigner(accountAddress)` method.

sunwrobert:
Handling this in #71

PR #71, opened 2022-07-07
feat: add support for custom signer

### LINKED ISSUE #48: Use ethers.Signer abstract class instead of providers.JsonRpcProvider
Requesting you change the Seaport `provider` constructor argument to the abstract class `ethers.Signer`. This would give developers much more flexibility in using this library. For example I could inject the gnosis safe adapter https://www.npmjs.com/package/@gnosis.pm/safe-ethers-adapters?activeTab=readme to easily interop with gnosis safe.

### LINKED ISSUE #62: Unknown Account error message
`Error: processing response error (body="{\"jsonrpc\":\"2.0\",\"id\":49,\"error\":{\"code\":-32000,\"message\":\"unknown account\"}}`
Code: 
```
const provider = new ethers.providers.JsonRpcProvider(customUrl, "mainnet");
const seaport = new Seaport(provider);
const { executeAllActions, actions } = await seaport.createOrder(
      {
        offer: [
          {
            itemType: 2,
            token, 
            identifier: tokenId,
            amount: "1",
            endAmount: "1",
          },
        ],
        consideration: [
          {
            itemType: 0,
            identifier: "0",
            amount: "9250000000000000",
            token: "0x0000000000000000000000000000000000000000",
            endAmount: "9250000000000000",
            recipient: offerer,
          },
          {
            itemType: 0,
            identifier: "0",
            token: "0x0000000000000000000000000000000000000000",
            amount: "250000000000000",
            endAmount: "250000000000000",
            recipient: "0x8De9C5A032463C561423387a9648c5C7BCC5BC90",
          },
          {
            itemType: 0,
            identifier: "0",
            token: "0x0000000000000000000000000000000000000000",
            amount: "500000000000000",
            endAmount: "500000000000000",
            recipient: "0x53fa2013286C57114EfF51D6A27E09e46Acc5f89",
          },
        ],
      },
      offerer
    );

    const order = await executeAllActions();
```

I think that issues comes from not having the account private key when creating the message.

Is there a way to provide a signer to seaport that can be used to sign. I have an environment where I cannot expect my users to have Meta Mask installed.

### LINKED ISSUE #70: Default provider confusion from the documentation
👋 Hey guys,

I am trying to figure out how to use this library instead of the opensea-js mostly of this library support of bulk purchasing of tokens.

No matter how I try to figure out what providers to use nothing works and documentation suggests to use

```typescript
const provider = ethers.getDefaultProvider();
```

Which is seemingly incorrect and doesn't work out of the box, as the library works only with `JsonRpcProvider` and `getDefaultProvider` exports `BaseProvider` that is missing `getSigner` method.

https://github.com/ProjectOpenSea/seaport-js/blob/5952c3cf162c7ab657a0ec455140df9a7f62ee60/src/seaport.ts#L67-L69

So trying to understand if I am doing something that I am not supposed to do, or we should really update the docs with more relevant example.

Thanks.

sunwrobert:
## Motivation
Give more flexibility to users that aren't using Seaport solely through browser providers. Allow them to provide a custom signer instead.

Fixes #48 
Fixes #62 
Fixes #70 

## Solution
Have the constructor take in a provider or a signer

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,10 +34,39 @@ npm install --save @opensea/seaport-js
 
 Instantiate your instance of seaport using your ethers provider:
 
+### Examples
+
+#### Through a browser provider (i.e. Metamask)
+
 ```js
 import { Seaport } from "@opensea/seaport-js";
+import { ethers } from "ethers";
 
-const provider = ethers.getDefaultProvider();
+const provider = new ethers.providers.Web3Provider(window.ethereum);
+
+const seaport = new Seaport(provider);
+```
+
+#### Through an RPC Provider (i.e. Alchemy)
+
+```js
+import { Seaport } from "@opensea/seaport-js";
+import { ethers } from "ethers";
+
+const provider = new ethers.provider.JsonRpcProvider(
+  "https://<network>.alchemyapi.io/v2/YOUR-API-KEY"
+);
+
+const seaport = new Seaport(provider);
+```
+
+#### With custom signer
+
+```js
+import { Seaport } from "@opensea/seaport-js";
+import { ethers } from "ethers";
+
+const signer = new ethers.Wallet("YOUR_PK");
 
 const seaport = new Seaport(provider);
 ```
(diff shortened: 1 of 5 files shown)


KatFishSnake (approved):
lgtm, works on my side, thank you 🙏 for the work

PR #72, opened 2022-07-07
separate file for cancel tests

arr00:


arr00:
> can we also add a test for cancelAll/incrementCounter?
> 
> can be in the same file

`bulkCancelOrders` uses the incrementCounter function on seaport

sunwrobert:
> > can we also add a test for cancelAll/incrementCounter?
> > can be in the same file
> 
> `bulkCancelOrders` uses the incrementCounter function on seaport

ah, didn't see that. LGTM

PR #76, opened 2022-07-08
fix: set signer in Seaport constructor

isvidler:
Fix bug where signer was not being set in constructor.

## Motivation
The Seaport constructor now accepts a signer, but doesn't actually set the `signer` class variable in the constructor, causing `_getSigner` to throw an error.

## Solution
Solution is to set `signer` in the constructor.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@opensea/seaport-js",
-  "version": "1.0.3",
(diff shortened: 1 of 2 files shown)


sunwrobert (approved):
ah woops, thanks for this.

sunwrobert:
@isvidler can you fix prettier?

PR #112, opened 2022-09-07
Add domain hash to random salt

stephankmin:
## Motivation
In order to attribute orders to their respective domains for analytics purposes, allow users to supply an optional `string domain` param when creating orders. The domain hash is included in the first four bytes of the arbitrary order salt. If no domain is supplied, first four bytes are empty (0x00000000).

## Solution
`generateRandomSaltWithDomain` returns a random salt with the supplied domain hashed and added to the first four bytes.


sunwrobert (approved):
lgtm. A bit awkward that there are two params now but they don't interact how one might expect when passing in both

0age (approved):

PR #117, opened 2022-09-23
Add hash of passed in domain to end of calldata

stephankmin:
## Motivation
Provide an on-chain method for attributing fulfillment of a given Seaport order to a specific domain

## Solution
Allow developers to pass in an optional string `domain`, hash the domain in `fulfill.ts` to create a `tag`, and append the tag to the end of calldata for `fulfillment`, `validate`, and `cancel` transaction calls

## Additional Context
- Fulfillment methods in `seaport.ts` and `fulfill.ts` now accept an optional string `domain` param
- If a domain is passed in, it is hashed via `getTagFromDomain` and the tag is passed into the call to `getTransactionMethods`
- If no domain is passed in, `getTagFromDomain` is not called and the tag passed into `getTransactionMethods` defaults to an empty string
- Updated tests to verify:
      1.  No tag is created when domain is left empty
      2. `getTagFromDomain` works with various domains
      3. Tag is appended to end of fulfillment calldata

## Open Questions
1. Should tag creation be handled in `seaport.ts` and `fulfill.ts` or in `getTransactionMethods` itself in `usecase.ts`?
2. Please ensure the optional param handling looks as expected (domain needs to be passed from fulfillment methods in `seaport.ts`, to fulfillment methods in `fulfill.ts`, to calls to `getTransactionMethods` in `usecase.ts`. 
3. Sanity check on [regex](https://github.com/ProjectOpenSea/seaport-js/blob/stephan/attributions-calldata/src/utils/fulfill.ts#L853) for handling "0x" prefix

## TODO
1. Write domain registry contract and add helper util function to SDK for registering domain/warning user if their domain isn't registered
2. Update Seaport README

### DIFF
--- a/src/seaport.ts
+++ b/src/seaport.ts
@@ -416,13 +416,17 @@ export class Seaport {
    */
   public cancelOrders(
     orders: OrderComponents[],
-    accountAddress?: string
+    accountAddress?: string,
+    suffix?: string
   ): TransactionMethods<ContractMethodReturnType<SeaportContract, "cancel">> {
     const signer = this._getSigner(accountAddress);
 
-    return getTransactionMethods(this.contract.connect(signer), "cancel", [
-      orders,
-    ]);
+    return getTransactionMethods(
+      this.contract.connect(signer),
+      "cancel",
+      [orders],
+      suffix
+    );
   }
 
   /**
@@ -431,7 +435,8 @@ export class Seaport {
    * @returns the set of transaction methods that can be used
    */
   public bulkCancelOrders(
-    offerer?: string
+    offerer?: string,
+    suffix?: string
   ): TransactionMethods<
     ContractMethodReturnType<SeaportContract, "incrementCounter">
   > {
@@ -440,7 +445,8 @@ export class Seaport {
     return getTransactionMethods(
       this.contract.connect(signer),
       "incrementCounter",
-      []
+      [],
+      suffix
     );
   }
 
@@ -453,13 +459,17 @@ export class Seaport {
    */
   public validate(
     orders: Order[],
-    accountAddress?: string
+    accountAddress?: string,
+    suffix?: string
   ): TransactionMethods<ContractMethodReturnType<SeaportContract, "validate">> {
     const signer = this._getSigner(accountAddress);
 
-    return getTransactionMethods(this.contract.connect(signer), "validate", [
-      orders,
-    ]);
+    return getTransactionMethods(
+      this.contract.connect(signer),
+      "validate",
+      [orders],
+      suffix
+    );
   }
 
   /**
--- a/src/seaport.ts
+++ b/src/seaport.ts
@@ -909,29 +909,30 @@ export class Seaport {
    * @param input.accountAddress Optional address for which to match the order with
    * @returns set of transaction methods for matching orders
    */
-  public matchOrders(
-    {
-      orders,
-      fulfillments,
-      overrides,
-      accountAddress,
-    }: {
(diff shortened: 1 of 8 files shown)


src/seaport.ts
```diff
@@ -626,6 +643,7 @@ export class Seaport {
     accountAddress,
     conduitKey = this.defaultConduitKey,
     recipientAddress = ethers.constants.AddressZero,
+    domain = "",
```
stephankmin, line 646: get an error when attempting to set this as optional param

src/seaport.ts
```diff
@@ -890,20 +914,25 @@ export class Seaport {
...
     ContractMethodReturnType<SeaportContract, "matchOrders">
   > {
     const signer = this._getSigner(accountAddress);
 
-    return getTransactionMethods(this.contract.connect(signer), "matchOrders", [
-      orders,
-      fulfillments,
-      overrides,
-    ]);
+    const tag = domain ? getTagFromDomain(domain) : "";
```
stephankmin, line 929: This ternary feels weird since the default value for domain is an empty string anyways

src/utils/fulfill.ts
```diff
@@ -829,3 +846,7 @@ export const getAdvancedOrderNumeratorDenominator = (
 
   return { numerator, denominator };
 };
+
+export const getTagFromDomain = (domain: string) => {
+  return keccak256(toUtf8Bytes(domain)).slice(0, 10).replace(/^(0x)/, "");
```
stephankmin, line 851: Does this regex look right?

src/utils/usecase.ts
```diff
@@ -85,17 +96,13 @@ export const getTransactionMethods = <
...
     },
-    transact: (overrides?: Overrides) => {
+    transact: async (overrides?: Overrides) => {
       const mergedOverrides = { ...initialOverrides, ...overrides };
 
-      return contract[method as string](...args, mergedOverrides);
-    },
-    buildTransaction: (overrides?: Overrides) => {
-      const mergedOverrides = { ...initialOverrides, ...overrides };
+      const data = await buildTransaction(mergedOverrides);
```
stephankmin, line 102: Call `transact` with the appended calldata generated in `buildTransaction`

src/utils/usecase.ts
```diff
@@ -70,6 +72,15 @@ export const getTransactionMethods = <
     args.pop();
   }
 
+  const buildTransaction = async (overrides?: Overrides) => {
+    const mergedOverrides = { ...initialOverrides, ...overrides };
+    const populatedTransaction = await contract.populateTransaction[
+      method as string
+    ](...[...args, mergedOverrides]);
+    populatedTransaction.data = populatedTransaction.data + tag;
```
stephankmin, line 80: Should the tag be created here (with the domain being passed into `getTransactionMethods`)?

src/utils/fulfill.ts
ryanio: you don't really need the regex, just do `.slice(2, 10)`

src/seaport.ts
ryanio: seems fine, the ternary operator is typically used in this way

src/utils/usecase.ts
ryanio: that would dedupe more of the `getTagFromDomain` uses that are passed in so I don't see why not. also `domain` seems more self explanatory than `tag` so that would be another benefit.

sunwrobert (commented):
approach lgtm

jtaaa (approved):

PR #124, opened 2022-10-04
Add getter/setter methods for Domain Registry contract

stephankmin:


src/seaport.ts
```diff
@@ -932,4 +943,27 @@ export class Seaport {
...
+    const signer = this._getSigner(accountAddress);
+
+    return getTransactionMethods(
+      this.domainRegistry.connect(signer),
+      "setDomain",
+      [domain]
+    );
+  }
+
+  public getDomain(tag: string, index: number): Promise<string> {
```
0age, line 962: Should include a getTotalDomains here

src/seaport.ts
```diff
@@ -932,4 +943,27 @@ export class Seaport {
...
+    return this.domainRegistry.getDomain(tag, index);
+  }
+
+  public async getDomains(tag: string): Promise<string[]> {
+    return this.domainRegistry.getDomains(tag);
```
0age, line 967: For this one, I'd say try/catch it and if it fails then fall back to getTotalDomains + map(getDomain)

0age (changes requested):
Small change to the getter proposed

0age (approved):

src/seaport.ts
```diff
@@ -932,4 +944,54 @@ export class Seaport {
       domain
     );
   }
+
+  public setDomain(
+    domain: string,
+    accountAddress: string
```
sunwrobert, line 950: this can be optional?

src/seaport.ts
```diff
@@ -932,4 +944,54 @@ export class Seaport {
...
+  public async getDomains(
+    tag: string,
+    shouldThrow?: boolean
+  ): Promise<string[]> {
+    try {
+      if (shouldThrow) {
+        throw Error;
+      }
+
+      this.domainRegistry.getDomains(tag);
```
sunwrobert, line 980: why don't we just return here? so we don't need to return below.

sunwrobert (approved):

PR #147, opened 2022-12-14
Support custom criteria in createOrder

xtools-at:
## Motivation

Seaport-js supports creating orders for NFTs using either a single token or a list of token ids. While this is convenient and saves us from messing with merkle trees ourselves, e.g. collection-wide offers for several 100k NFTs become problematic to execute and crippling user experience, since we're creating and processing an endless list of ids on the client side.
For this particular use case, you may want to compute a merkle root on the server side to spare the browser the work.

## Solution

this change enables passing in a pre-computed merkle root into the `createOrder` call for ERC721 and ERC1155, using a new `criteria` param (as opposed to using `identifier` or `identifiers`)

### DIFF
--- a/src/__tests__/create-order.spec.ts
+++ b/src/__tests__/create-order.spec.ts
@@ -6,6 +6,7 @@ import { ItemType, MAX_INT, NO_CONDUIT, OrderType } from "../constants";
 import { ApprovalAction, CreateOrderAction } from "../types";
 import { generateRandomSalt } from "../utils/order";
 import { describeWithFixture } from "./utils/setup";
+import { MerkleTree } from "../utils/merkletree";
 
 describeWithFixture("As a user I want to create an order", (fixture) => {
   it("should create the order after setting needed approvals", async () => {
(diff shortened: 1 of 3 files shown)


0age:
This makes sense to me, though it's worth flagging that there's a degree of trust that the server is in fact returning the correct criteria; I wonder if the client could still take the tokenIds along with the criteria root, then randomly sample a much smaller number of them and just derive proofs for those and ensure they are present in the supplied root?

(also looks like tests are not currently passing?)

nfts4f4fff:
I've been trying to figure this merkle root thing out. The tests aren't working rly when it comes to criteria based items. I can't believe that this issue is still not fixed. No docs explaining how to do so. How are we supposed to make a merkle proof of collection with 100k users. When I create a array from 1-10k let's say. And then get a merkle proof from it. The contract returns invalid proof ! I cant build my app bevause of that now almost for a week @0age  @xtools-at @sunwrobert

xtools-at:
superseded by #196 , i've rebased the changes and tests on seaport-js 1.1.0

PR #155, opened 2023-01-21
Add seaport v1.2 and bulk order signatures

ryanio:
This PR adds Seaport v1.2 and bulk order signatures. Special thanks to @d1ll0n for the great typescript utils ported from the Seaport repo.

TLDR on bulk sigs: The order signature field includes the BulkOrder signature, order index, and proof.

I deduplicated some of the order prep code between `createOrder()` and the new `createBulkOrders()` into a new internal function `_formatOrder()`, to help future maintainability.

There was also both a `package-lock.json` and a `yarn.lock`. Since the readme says to `npm install` rather than use yarn, I kept `package-lock.json` and removed `yarn.lock` and added it to gitignore so it won't accidentally be added in the future.


src/seaport.ts
```diff
@@ -287,11 +363,10 @@ export class Seaport {
       salt ||
       (domain ? generateRandomSaltWithDomain(domain) : generateRandomSalt());
 
-    const orderParameters: OrderParameters = {
+    const orderComponents: OrderComponents = {
       offerer,
       zone,
-      // TODO: Placeholder
-      zoneHash: formatBytes32String(resolvedCounter.toString()),
+      zoneHash: ethers.constants.HashZero,
```
ryanio, line 369: i'm not sure why the zoneHash was being set to the counter here so i changed it to zero hash

.gitignore
```diff
@@ -4,6 +4,7 @@ coverage
 coverage.json
 typechain
 lib
+yarn.lock
```
sunwrobert, line 7: 👍🏼

src/utils/eip712/defaults.ts
```diff
@@ -0,0 +1,113 @@
+/* eslint-disable no-dupe-class-members */
+/* eslint-disable no-unused-vars */
```
sunwrobert, line 2: these still needed?

sunwrobert (approved):
lgtm generally

src/utils/eip712/defaults.ts
ryanio: didnt need no-unused-vars, and i was able to reduce the scope of no-dupe-class-members, pushing a commit for that now 👍

PR #171, opened 2023-02-20
Update to v1.4

ryanio:
Seaport 1.2 and 1.3 were deprecated for small issues, so we will be using 1.4


package.json
```diff
@@ -92,7 +92,7 @@
     "prettier-package-json": "^2.6.3",
     "prettier-plugin-solidity": "^1.0.0-beta.13",
     "seaport": "ProjectOpenSea/seaport#76e508657888e3ec84039e1da269c822826991ef",
-    "seaport_v1_2": "ProjectOpenSea/seaport#1.2",
+    "seaport_v1_2": "ProjectOpenSea/seaport#1.4",
```
0age, line 95: This still calls it seaport_v1_2

ryanio: thanks good catch, updated in https://github.com/ProjectOpenSea/seaport-js/pull/171/commits/347baf9d637be63b3cb5d24d18aa71d377337eac

0age (approved):

PR #173, opened 2023-02-24
bump version to 1.1.0, change seaport version to 1.4 by default

ryanio:


ryanio:
we will change the default next week, and do a patch bump for now with the 1.4 functionality

PR #178, opened 2023-03-09
Update BulkOrderElements type

0xShisui:
## Motivation

The `Eip712MerkleTree` class has been built as a generic class, however the `BulkOrderElements` type used in the `getDataToSign` method hardcode a reference to `OrderComponents`.

## Solution

Make `BulkOrderElements` type generic (`BulkOrderElements<T>`) and use `BaseType` in the `getDataToSign` function to return `BulkOrderElements<BaseType>`.


ryanio (approved):
lgtm, thanks!

PR #180, opened 2023-03-09
Change the default configuration to Seaport 1.4

JoshuaSchmidt-OpenSea:
The API no longer accepts Seaport 1.4 so the default configuration should now be changed to use Seaport 1.4

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@opensea/seaport-js",
-  "version": "1.1.10",
+  "version": "1.1.0",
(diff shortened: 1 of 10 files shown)


montasaurus (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@opensea/seaport-js",
-  "version": "1.0.10",
+  "version": "1.1.10",
```
ryanio, line 3: Version would be 1.1.0

ryanio (changes requested):

ryanio (approved):

PR #196, opened 2023-03-31
Support custom criteria in createOrder [rebased]

xtools-at:
## Motivation

Seaport-js supports creating orders for NFTs using either a single token or a list of token ids. While this is convenient and saves us from messing with merkle trees ourselves, e.g. collection-wide offers for several 100k NFTs become problematic to execute and crippling user experience, since we're creating and processing an endless list of ids on the client side.
For this particular use case, you may want to compute a merkle root on the server side to spare the browser the work.

This PR was rebased on Seaport 1.1.0 and supersedes #147  

## Solution

this change enables passing in a pre-computed merkle root to the `createOrder` call for ERC721 and ERC1155, using a new `criteria` param (as opposed to using `identifier` or `identifiers`)

### DIFF
--- a/src/__tests__/criteria-based.spec.ts
+++ b/src/__tests__/criteria-based.spec.ts
@@ -564,6 +564,96 @@ describeWithFixture(
             expect(fulfillStandardOrderSpy).calledTwice;
           });
 
+          it("ERC721 <=> ETH (custom merkle root)", async () => {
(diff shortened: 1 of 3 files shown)


sunwrobert (approved):

sunwrobert:
Thanks!

PR #190, opened 2023-03-17
1155 partiall fill rounding issue

wooglie:
## Motivation

Wrote a test showcasing this issue https://github.com/ProjectOpenSea/seaport-js/issues/141

Thrown error
```
[...]

Issue #185, opened 2023-03-14
There is a bug in the mapOrderAmountsFromUnitsToFill method, when it encounters a case like 10000/3.


tercel:
The mapOrderAmountsFromUnitsToFill function means to calculate the amount needed for partial transactions.
Using multiplyBasisPoints  to calculate amount is inaccuracy, 
// i.e if totalFilled is 3 and totalSize is 4, there are 1 / 4 order amounts left to fill.
  const basisPoints = totalSize.sub(totalFilled).mul(ONE_HUNDRED_PERCENT_BP).div(totalSize);
for example, totalSize=3, totalFilled=2, ONE_HUNDRED_PERCENT_BP=10000 
basisPoints = 333, can't get rid of it all.

It should be modified as follows:
`+ const multiplyDivision = (amount: BigNumberish, numerator: BigNumberish, denominator: BigNumberish) =>`
`+   BigNumber.from(amount).mul(BigNumber.from(numerator)).div(denominator);`

`- startAmount: multiplyBasisPoints(item.endAmount, basisPoints).toString(),`
`+ startAmount: multiplyDivision(item.startAmount, unitsToFillBn, totalSize).toString(),`
`- endAmount: multiplyBasisPoints(item.endAmount, basisPoints).toString(),`
`+ endAmount: multiplyDivision(item.endAmount, unitsToFillBn, totalSize).toString(),`

xtools-at:
together with @wooglie we've tested your proposed solution and built a PR from it, thanks for that :) #197

PR #145, opened 2022-12-01
Demonstrate rounding error through unit test

0xMasayoshi:
## Motivation

This PR creates a unit test which demonstrates a rounding error in `mapOrderAmountsFromUnitsToFill` as explained in #141.

Issue #141, opened 2022-11-19
Rounding error in `mapOrderAmountsFromUnitsToFill`


0xMasayoshi:
There is a potential rounding error when calculating offer/consideration amounts in `mapOrderAmountsFromUnitsToFill`.
&nbsp;  
Consider the following example:
- Alice creates a partially fillable order, listing 3 of some ERC1155 token for 1ETH each
- Bob tries to buy 1 of these tokens by calling `seaport.fulfillOrder`
- In this scenario, `seaport.fulfillOrder` will execute `fulfillAdvancedOrder` on the Seaport contract with a value of 0.9999ETH, resulting in InsufficientEtherSupplied error

---

Let's examine why this is happening using the following snippets from `mapOrderAmountsFromUnitsToFill` as reference:
```
const unitsToFillBasisPoints = unitsToFillBn
    .mul(ONE_HUNDRED_PERCENT_BP)
    .div(maxUnits);
```
&&
```
const multiplyBasisPoints = (amount: BigNumberish, basisPoints: BigNumberish) =>
  BigNumber.from(amount)
    .mul(BigNumber.from(basisPoints))
    .div(ONE_HUNDRED_PERCENT_BP);
```
&&
```
startAmount: multiplyBasisPoints(
    item.startAmount,
    basisPoints
).toString(),
```

When Bob executes `seaport.fulfillOrder` the following calculations are made:
- `unitsToFillBasisPoints = 1 * 10_000 / 3 = 3333.3333333333333333 = 3333 (due to BigNumber truncation)`
- `startAmount: 3E18 * 3333 / 10_000 = 0.9999E17 = 0.9999ETH`

Because BigNumber does not support decimals, the precision of `unitsToFillBasisPoints` is off, resulting in a rounding error when calculating the consideration's start and end amounts.

adeebabdulsalam:
what is maxUnits ? is it the length of offer / consideration or startAmount/endAmount.
I faced this error when my offer is partially fillable ERC1155 token (startAmount: around 100)
unitsToFill =1 , and consideration length =3 (payment,commission and royalty).

I hope there is a fix soon

luisantoniocrag:
Is there some temporally solution?

xtools-at:
created PR #197 together with @wooglie to solve this issue, please read the limitations stated in the description

PR #197, opened 2023-03-31
Prevent rounding issue with partial fills

xtools-at:
## Motivation

A rounding error is plaguing partial fulfilment with seaport-js, as discussed in e.g. #132 and #141 , and demonstrated via #145 and #190 . In #185 , user @tercel proposed a solution, which @wooglie turned into a proof of concept [here](https://github.com/wooglie/seaport-js/pull/1/files), that seems to tick all the boxes.

## Solution

I've ported @tercel 's solution to a clean slate of 1.1.0, together with some slight changes to @0xMasayoshi 's unit tests from #145. I did some testing with various values (different prices and precisions, and fee-basisPoints) and it seems to work fine now, yet there are some limitations to the proposed solution.

## Limitations
There's a lot of back and forth going on in the discussions, read this carefully to understand the limitations to handle the use case properly.

_TL/DR: create your orders by multiplying the number of nfts with the price per nft, don't set an arbitrary price for the whole bundle if you want to do partial fills. if your orders uses seaport-js's **fees** implementation, the price per nft shouldn't have more than 12 decimals (for currencies with 18 decimals)_ 

In detail:
- "10 ETH / 3 NFT doesn't work" - well yeah, even with floating points you'd end up with an infinitely small amount of inaccuracy (3.33333333...), and this fix won't solve that particular case. you have to *create* your order by multiplying the **amount of NFTs** with the target **sale price per NFT** already. so instead of saying "i want 10 ETH for 3 NFTs", you have to approach it as "i want 3.333 ETH for each NFT, for up to 3 units", if you want partial fills - and this circumvents the problem altogether by having clear divisibility (see [here](https://github.com/ProjectOpenSea/seaport-js/pull/197/files#diff-dca5b1ad9aeafdacac4c558083e92ba71349fa114df87ae7633f993c2199af9fR171-R176) for reference).
- when we're applying fees, things can get a bit murky still. we can again run into a situation where the resulting fees are not divisible by the number of units, because the basis points may be something arbitrary set by the user and out of our control. this fix solves the issue to an extent, but you'd have to limit the number of decimals of your sale price. Using e.g. ETH (18 decimals), your price can have up to 12 decimals to allow for enough "room" for the calculation to be safe. it _may_ work if with a higher number of decimals too for some nice round values, but you might be starting to run into edge cases when e.g. someone sets their NFT royalty to an odd number like 2.43% (see [here](https://github.com/ProjectOpenSea/seaport-js/pull/197/files#diff-dca5b1ad9aeafdacac4c558083e92ba71349fa114df87ae7633f993c2199af9fR153-R179) for reference).
  - OPEN: how are currencies with less decimals affected? e.g. USDC has 6 only decimals, how precise can we get --> to verify

### DIFF
--- a/src/__tests__/partial-fulfill.spec.ts
+++ b/src/__tests__/partial-fulfill.spec.ts
@@ -147,6 +147,79 @@ describeWithFixture(
           expect(fulfillStandardOrderSpy).calledOnce;
         });
 
+        it("ERC1155 <=> ETH doesn't fail due to rounding error", async () => {
+          const { seaport, testErc1155 } = fixture;
+
+          // broke out key params to make testing different values easier:
+          const unitsForSale = "3";
+          // a unit sale price with a precision up to 12 decimals seems to work in combination with *any* fee (for tokens with 18 decimals).
+          // if *no* fees are involved, 18 decimals may work fine. in-between your mileage may vary, depending on the value of `basisPoints`.
+          const pricePerUnit = "3.1415926535897";
+          const basisPoints = 247;
+
+          // maker creates partially fillable listing with amount of 3
+          standardCreateOrderInput.offer = [
+            {
+              itemType: ItemType.ERC1155,
+              token: testErc1155.address,
+              amount: unitsForSale,
+              identifier: nftId,
+            },
+          ];
+
+          // calculate total price (price per unit * units for sale) and fees
+          standardCreateOrderInput.consideration = [
+            {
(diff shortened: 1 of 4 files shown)


sunwrobert (approved):
LGTM

xtools-at:
UPDATE (also added in description above): for USDC, the sale price can have **max two decimals** only to be calculated safely when using fees and partial fills. will add a unit test in a new PR to confirm.

PR #198, opened 2023-04-03
Added unit test for USDC + rounding error (ERC20 tokens with less than 18 decimals)

xtools-at:
## Motivation

In #197 the rounding issue with partial fills was solved, with the limitation that the sale price is only allowed to have a certain amount of decimal places for the maths to safely check out. At the time of authoring that PR, it was still unclear how ERC20 tokens with less than 18 decimals are affected by the limitations laid out in #197.

## Solution

This PR adds an additional unit test with a mocked USDC ERC20 token, to prove the eligibility of the solution as well as the further limitations.

## Result

Prices for orders paid in ERC20 tokens with 6 decimals (like USDC) must not have more than two decimals, when used together with seaport-js' **fees** feature _and_ **partial fills** (i.e. `1.23` per NFT is ok, `1.234` may break for certain fee-basisPoints values).
When _not_ using **fees**, this limitation doesn't apply, i.e. you can use the full range of six decimals safely.


ryanio (approved):
nice, lgtm, thanks!

Issue #221, opened 2023-04-24
ERC20 Problem when used as an offer (Problem with latest seaport-js version)


MiguelGonzales-WIV:
```
const { executeAllActions } = await seaport.createOrder(
{
	offer: [
		// buyer's offer
		{
			token: '0xc1d6ef502ac5410b3f3706beb6a0808131337fb6',
			amount: offerPrice,
		}
	],
	consideration: [
		// buyer's ask
		{
			itemType: ItemType.ERC721,
			token: smartContractAddress,
			identifier: tokenID,
			recipient: address, // recipient for token is the buyer
		},
	],
},
address
)
```
I recently updated my `@opensea/seaport-js` to the latest version and now my code doesn't work.

Using this block of code to try and create an offer for some NFT. Thing is, when I put my token in the offer, a Metamask error pops up, indicating "Invalid token value; should be exactly 64 hex digits long (u256)".

If I change the offer to properties that represent an ERC721 token, it works totally fine.
Also, removing `token` property from the object in offer array, it pushes through.

I need to be able to indicate which ERC20 token address I want as an offer, but it isn't working.

Hope to get some help!

HeptaneL:
You can manually approve the ERC20 token to the seaport contract, and then execute the createOrder function.

ryanio:
This should be resolved now with [v1.2.1](https://github.com/ProjectOpenSea/seaport-js/releases/tag/v1.2.1)

Issue #215, opened 2023-04-13
Approve ERC20 token problem with metamask


AnhNguyenDuc2:
function buildTransaction in src/utils/usecase.ts attach domain string using getTagFromDomain function and when dmain string is empy the domain still attached hashed empty string to calldata cause the trouble when calling approval from Metamask. 

Metamask throw : " Invalid token value; should be exactly 64 hex digits long (u256)"

since Metamask check the approval number string tobe exact 64 character.

MiguelGonzales-WIV:
Same problem!

gabspeck:
@MiguelGonzales-WIV if it's of interest to you, I managed to work around this by forking the library and adding a conditional around the code that appends the custom "domain" hash to the transaction data: https://github.com/ProjectOpenSea/seaport-js/commit/2250434643766ee52f778a61548c6064362bd737

MiguelGonzales-WIV:
Seems it being added to the transaction data for other methods (e.g. fulfillOrder, fulfillOrders, and matchOrders). Has this been working for you for createOrder, where nature of order is an offer (i.e. offer property value is an ERC20 token)?
```
createOrder({
	offer: [
		// buyer's offer
		{
			token: '0xc1d6ef502ac5410b3f3706beb6a0808131337fb6',
			amount: offerPrice,
		}
	],
        ...rest of code
```

createOrder only works for ERC721 for me.

ryanio:
This should be resolved now with [v1.2.1](https://github.com/ProjectOpenSea/seaport-js/releases/tag/v1.2.1)

PR #222, opened 2023-04-25
Update domain check before adding to txn data

MiguelGonzales-WIV:


ryanio:
thanks for the PR! is there a test for this already, with domain both empty "" and with an entry, so that we ensure both cases work? if not, do you mind adding? ty!

ryanio:
Related issues: https://github.com/ProjectOpenSea/seaport-js/issues/221 https://github.com/ProjectOpenSea/seaport-js/issues/215

ryanio (approved):

PR #128, opened 2022-10-11
Strip unnecessary async functions and replace with normal functions

peteremiljensen:
## Motivation

These three low-level functions are unnecessary async functions:

- `fulfillBasicOrder` in `src/utils/fulfill.ts`
- `fulfillStandardOrder` in `src/utils/fulfill.ts`
- `getApprovalActions` in `src/utils/approval.ts`

## Solution

This PR converts them to normal functions by stripping off the `async`-keyword.

PR #246, opened 2023-05-24
remove unneeded async fns, remove esModuleInterop, upgrade output to es6, bump package.json version to v2

ryanio:
Closes https://github.com/ProjectOpenSea/seaport-js/pull/128

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,11 +1,10 @@
 {
   "compilerOptions": {
     "outDir": "lib",
-    "target": "es5",
+    "target": "es6",
     "lib": ["dom", "esnext"],
-    "allowSyntheticDefaultImports": true,
-    "skipLibCheck": true,
     "forceConsistentCasingInFileNames": true,
+    "skipLibCheck": true,
     "noImplicitReturns": true,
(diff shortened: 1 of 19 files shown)


tsconfig.json
```diff
@@ -1,11 +1,10 @@
 {
   "compilerOptions": {
     "outDir": "lib",
-    "target": "es5",
+    "target": "es6",
```
JoshuaSchmidt-OpenSea, line 4: Any desire to go further than this? Bigint was added in ES2020 but I don't know the downside of using a fairly new compiler target

ryanio: i think going to es6 is fine for now. we can do es2020 with ethers v6 + bigint upgrade

JoshuaSchmidt-OpenSea (approved):
Is the target version bump the reason for the major version upgrade?

ryanio:
actually the initial reason was changing the async signatures on some fns:
- fulfillBasicOrder in src/utils/fulfill.ts
- fulfillStandardOrder in src/utils/fulfill.ts
- getApprovalActions in src/utils/approval.ts

PR #250, opened 2023-05-26
Improve input type for salt

ryanio:
Changes input type for salt from string to BigNumberish to support more input formats, converts to value with `BigNumber.from(salt)`

```diff
---  salt: string;
+++  salt: BigNumberish;
```

This PR also moves the test folder out of src since we are including src for source maps now.


src/seaport.ts
```diff
@@ -739,7 +739,10 @@ export class Seaport {
             .slice(2)
             .padStart(64, "0"),
           orderComponents.zoneHash.slice(2),
-          orderComponents.salt.slice(2).padStart(64, "0"),
+          BigNumber.from(orderComponents.salt)
```
stuckinaboot, line 742: this will likely save others from issues with having salt represented improperly. Thx!

stuckinaboot (approved):

PR #286, opened 2023-06-14
Respect ERC1271

sosoxuc:
## Motivation
Allow ERC1271 orders

## Solution
Offerer can be different from signer (e.g. contract wallet address)


sosoxuc:
> i think we would also need to skip the `signOrder` step when 1271?

Do we?
An EOA can sign the order and then it is checked inside 1271 contract

```
    address public signer;

    function isValidSignature(bytes32 hash_, bytes calldata signature_) external override view returns (bytes4) {
        address signer_ = hash_.recover(signature_);
        if (signer_ == signer) {
            return 0x1626ba7e;
        }
        return 0x00000000;
    }
```

sosoxuc:
> nice thanks, do you mind also adding a 1271 test to ensure it works and doesn't break in the future?

Sure, will do

sosoxuc:
> nice thanks, do you mind also adding a 1271 test to ensure it works and doesn't break in the future?

Added tests but I have side effect — https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/ECDSA.sol#L70 will reject compacted sig

I can revert that change in the code if you do not think that makes sense

sosoxuc:
@ryanio

src/contracts/test/TestERC1271Wallet.sol
```diff
@@ -0,0 +1,71 @@
+//SPDX-License-Identifier: Unlicense
+pragma solidity >=0.8.7;
+
+import "hardhat/console.sol";
```
ryanio, line 4: do you mind removing this import

test/erc1271-order.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+import { ethers } from "hardhat";
+import { ItemType, MAX_INT } from "../src/constants";
+import { CreateOrderAction } from "../src/types";
+import { generateRandomSalt } from "../src/utils/order";
+import { describeWithFixture } from "./utils/setup";
+
+const OPENSEA_DOMAIN = "opensea.io";
+const OPENSEA_TAG = "360c6ebe";
+
+describeWithFixture(
```
ryanio, line 12: instead of a new test file, could you just add the test to the existing create-order.spec.ts?

sosoxuc:
@ryanio applied all changes

ryanio:
looks good, could you run `npm run prettier:fix` please

ryanio (approved):

PR #308, opened 2023-07-12
Allow `zoneHash` to be passed in `input` to `createOrder` and `createBulkOrders`

rombrom:
Since `zoneHash` is a mechanism for contracts utilizing the SeaPort infra to pass along arbitrary values it should be able to be passed as part of `input` in the order creation methods.

### DIFF
--- a/src/seaport.ts
+++ b/src/seaport.ts
@@ -296,6 +296,7 @@ export class Seaport {
     {
       conduitKey = this.defaultConduitKey,
       zone = ethers.constants.AddressZero,
+      zoneHash = ethers.constants.HashZero,
       startTime = Math.floor(Date.now() / 1000).toString(),
       endTime = MAX_INT.toString(),
       offer,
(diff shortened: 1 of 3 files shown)


ryanio:
Lgtm, thanks for opening! Do you mind adding a test to ensure that the value is passed through correctly to the resulting retuned order? Just so it won't break in the future in case code gets changed around.

rombrom:
> Lgtm, thanks for opening! Do you mind adding a test to ensure that the value is passed through correctly to the resulting retuned orders? Just so it won't break in the future in case code gets changed around.

Sure thing! Will add it later today!

rombrom:
Added a test case. Is this alright? Or do we require something more elaborate? I tested whether both `zone` and `zoneHash` are passed unaltered.

nazariyv:
@ryanio kindly bumping this

ryanio (approved):

ryanio:
@nazariyv released in [v2.0.5](https://github.com/ProjectOpenSea/seaport-js/releases/tag/v2.0.5)

@rombrom thank you for your contribution!

sofianeOuafir:
I love you, just what I needed

PR #310, opened 2023-07-17
Make `Seaport.getOrderHash()` a static method

rombrom:
Currently users need an instantiated `Seaport` to have access to `getOrderHash()` but there are use cases where one would want to get the `orderHash` outside of contexts where `Seaport` is instantiated. One such scenario would be storing orders in a database where modules could benefit from having static access to this method.

`getOrderHash()` doesn't utilize any instance internals so in changing this I've changed the method declaration to `public static getOrderHash()`. Since this would be a breaking change, I've re-added the instance method which now utilizes `Seaport.getOrderHash()` under the hood and issues a deprecation warning.

I've modified the `"should have the same order hash as on the contract"` test to check both the static and instance versions of `getOrderHash()`.

### DIFF
--- a/src/seaport.ts
+++ b/src/seaport.ts
@@ -640,8 +640,20 @@ export class Seaport {
   /**
    * Calculates the order hash of order components so we can forgo executing a request to the contract
    * This saves us RPC calls and latency.
+   * @deprecated Please use the static `Seaport.getOrderHash()` method.
    */
-  public getOrderHash = (orderComponents: OrderComponents): string => {
+  public getOrderHash(orderComponents: OrderComponents): string {
+    console.warn(
+      "seaport.getOrderHash() is deprecated in favor of the static Seaport.getOrderHash()",
+    );
+    return Seaport.getOrderHash(orderComponents);
+  }
+
+  /**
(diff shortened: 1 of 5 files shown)


ryanio:
seems reasonable, although in the future if the order hash calculation changes we do need to know which seaport version to calculate the hash for, which means we would need it to be instantiated, so perhaps this isn't a good future-proof solution

rombrom:
> seems reasonable, although in the future if the order hash calculation changes we do need to know which seaport version to calculate the hash for, which means we would need it to be instantiated, so perhaps this isn't a good future-proof solution

Hmm good point I hadn't considered. With `seaport-js` current library architecture in mind there really isn't a really solid pattern for exposing this statically then I think. Extending it to `Seaport.getOrderHash(parameters, version)` feels iffy. Would be cool if we could do something like `import { getOrderHash } from '@opensea/seaport-js/v1.5/utils';` but that would be a big departure from how this library is currently set up.

For our current use-cases this isn't a real issue so I'll close this to keep open PRs clean. lmk if you think this does have merit. I'll keep the changes around on my remote.

PR #340, opened 2023-08-17
Fix undefined error in tip parsing

frankisawesome:
## Motivation

The `fulfillOrders` flow when using a tip will hit this function and fungible (NATIVE) tokens will be mapped to have missing `identifier` field causing errors downstream.

I removed the errornous comment and implemented the fix but due to limited capacity I didn't bother with tests :)


ryanio:
Thanks! Can you use the nullish coalescing operator `??` instead of `||`?

frankisawesome:
@ryanio done, thank you

ryanio (approved):

PR #346, opened 2023-09-23
Patched Fix Improperly Controlled Modification of Object PrototypePollution leads to module code execution

iot-defcon:
## Sumarry 🐛
**CWE-1321 :** Improperly Controlled Modification of Object Prototype Attributes ('Prototype Pollution') The product receives input from an upstream component that specifies attributes that are to be initialized or updated in an object, but it does not properly control modifications of attributes of the object prototype. By freezing the object prototype first (for example, Object.freeze(Object.prototype)), modification of the prototype becomes impossible. By blocking modifications of attributes that resolve to object prototype, such as proto or prototype, this weakness can be mitigated. When handling untrusted objects, validating using a schema can be used. By using an object without prototypes (via Object.create(null) ), adding object prototype attributes by accessing the prototype via the special attributes becomes impossible, mitigating this weakness. Map can be used instead of objects in most cases. If Map methods are used instead of object attributes, it is not possible to access the object prototype or modify it. Prototype pollution by merging objects. Prototype pollution by setting default values to object attributes recursively. Prototype pollution by merging objects recursively. Prototype pollution by setting object attributes based on dot-separated path.

## Description 
The project of `ProjectOpenSea/seaport-js` used flat module could allow a remote attacker to execute arbitrary code on the system, caused by a prototype pollution flaw in the unflatten function in index.js. By adding or modifying properties of Object.prototype using a `__proto__` or constructor payload, an attacker could exploit this vulnerability to execute arbitrary code or cause a denial of service condition on the system.

**Proof On Concepts:**
```js
    let recipient = result

    while (key2 !== undefined) {
const type = Object.prototype.toString.call(recipient[key1])
      const isobject = (
        type === '[object Object]' ||
        type === '[object Array]'
      )
      // do not write over falsey, non-undefined values if overwrite is false
      if (!overwrite && !isobject && typeof recipient[key1] !== 'undefined') {
        return
      }
      if ((overwrite && !isobject) || (!overwrite && recipient[key1] == null)) {
        recipient[key1] = (
          typeof key2 === 'number' &&
          !opts.object ? [] : {}
        )
      }
      recipient = recipient[key1]
      if (split.length > 0) {
        key1 = getkey(split.shift())
        key2 = getkey(split[0])
      }
```
```diff
+      if (key1 === '__proto__') {
+        return
+      }
```
```js
var unflatten = require('flat').unflatten;
unflatten({ 'proto.polluted': true });
```
```
console.log(polluted); // true
```
**Impact:**
CWE-1321
`CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H`
9.0 Score
`0.2 %`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@opensea/seaport-js",
-  "version": "2.0.6",
+  "version": "2.0.7",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@opensea/seaport-js",
-      "version": "2.0.6",
+      "version": "2.0.7",
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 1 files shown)


ryanio:
can you explain what this PR is changing

xtools-at:
i suppose it prevents an XSS attack pattern

Issue #82, opened 2022-07-19
Updating gas fees for transaction


mikefogg:
Hey there! I totally could be missing something, but it doesn't look like there's a way as of now to update the gas fees (priority fee, mainly) for a certain transaction before fulfilling it.

It looks like it's _technically_ built in a bit by allowing updates to `payableOverrides` to make their way into the `getTransactionMethods` function, but there's currently no way to update `payableOverrides` via the .fulfillX methods :) 

Is there a downside to allowing users to pass through something like:

```
seaport.fulfillOrder({
  order,
  accountAddress,
  payableOverrides: {
    gas: '...',
    maxPriorityFeePerGas: '...',
    maxFeePerGas: '...',
  }
})
```

sunwrobert:
Hmm.. it might make more sense to pass that into `executeAllActions` then, no? You need to either get the `actions` and call `transactionMethods` anyways or just call `executeAllActions`.

mikefogg:
That's a really good point!

The benefit is that we'd be able to set it to pay a bit more gas on the order creation if we want to, just a bit more flexibility.

Happy to whip up a PR for review if that's something that you think makes sense to add :)

sunwrobert:
Yeah please do!

mikefogg:
Didn't forget about this! Stuck on a project this week, should have a PR next week :)

jdom1824:
if we can manipulate the gas price?  or at least set a range?

PR #363, opened 2023-09-29
Add payableOverrides to the fulfillOrder method.

tenthirtyone:
Adds an optional `overrides` parameter to `fulfillOrder`. Has the same functionality as `overrides` in `matchOrders` except it will ignore `value` if provided as that is calculated elsewhere and based on the order.

## Motivation

I wanted to be able to set my own gasLimit in fulfillOrder and saw someone else had previously made an issue but never finished so I thought I would go ahead.

https://github.com/ProjectOpenSea/seaport-js/issues/82

## Solution

``` js
const { actions } = await seaport.fulfillOrder({
  order,
  accountAddress: fulfiller.address,
  domain: OPENSEA_DOMAIN,
  overrides: { gasLimit: overrideGasLimit }, // new feature
});
```

### DIFF
--- a/src/seaport.ts
+++ b/src/seaport.ts
@@ -780,6 +780,7 @@ export class Seaport {
    *                               Defaults to the zero address which means the offer goes to the fulfiller
    * @param input.domain optional domain to be hashed and appended to calldata
    * @param input.exactApproval optional boolean to indicate whether the approval should be exact or not
+   * * @param input.overrides any overrides the client wants, will ignore value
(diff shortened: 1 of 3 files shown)


ryanio:
thanks for opening and adding tests :) we can definitely get this in and also add to opensea-js methods that send out txs

ryanio:
thinking if we should add these overrides in a standard way to all the other functions that send txs, other than just fulfillOrder 🤔

ryanio (approved):

PR #460, opened 2024-01-04
Remove incorrect Signer type

KenLR:
## Motivation

This pull request replaces the custom `Signer` type with the standard type from ethers. This update is due to changes in the ethers v6 library, specifically the deprecation and subsequent removal of the `_signTypedData` function. The existing custom `Signer` type, which extends `ethers.Signer` with `_signTypedData`, is now obsolete and incorrect.

## Solution

- Remove the custom type and related ones from `types.ts`.
- Update the `seaport.ts` import to use the standard `Signer` definition.

### DIFF
--- a/src/seaport.ts
+++ b/src/seaport.ts
@@ -6,6 +6,7 @@ import {
   JsonRpcProvider,
   Provider,
   JsonRpcSigner,
+  Signer,
 } from "ethers";
 import {
   SEAPORT_CONTRACT_NAME,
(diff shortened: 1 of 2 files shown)


ryanio:
nice catch, thank you!

ryanio (approved):

ryanio:
released in [v3.0.2](https://github.com/ProjectOpenSea/seaport-js/releases/tag/v3.0.2)

Issue #558, opened 2024-05-29
Scale tip amounts to partial fills when using fulfillOrders (fulfillAvailableAdvancedOrders)


frankisawesome:
### Component

API

### Have you ensured that all of these are up to date?

- [X] seaport-js
- [X] Node (minimum v16)

### What version of seaport-js are you on?

latest

### What function is the bug in?

fulfillOrders

### Operating System

None

### Describe the bug

When calling fulfillOrders with a partial order fulfillment and using a tip, the tip amount is not scaled to the fulfillment proportion. This behaviour was fixed in fulfillOrder.

I intend to have a PR shortly fixing this 👍

PR #559, opened 2024-05-29
(fix) Fix tip amount scaling in fulfillOrders

frankisawesome:
## Motivation

Fixes https://github.com/ProjectOpenSea/seaport-js/issues/558

### DIFF
--- a/src/seaport.ts
+++ b/src/seaport.ts
@@ -36,6 +36,7 @@ import type {
   ApprovalAction,
   CreateBulkOrdersAction,
   SeaportContract,
+  ConsiderationItem,
 } from "./types";
 import { getApprovalActions } from "./utils/approval";
 import {
@@ -52,8 +53,9 @@ import {
   shouldUseBasicFulfill,
   validateAndSanitizeFromOrderStatus,
 } from "./utils/fulfill";
-import { isCurrencyItem } from "./utils/item";
+import { getMaximumSizeForOrder, isCurrencyItem } from "./utils/item";
 import {
+  adjustTipsForPartialFills,
   areAllCurrenciesSame,
   deductFees,
   feeToConsiderationItem,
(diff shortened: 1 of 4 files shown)


src/seaport.ts
```diff
@@ -1029,24 +1031,38 @@ export class Seaport {
...
+              recipient: tip.recipient,
+            })) ?? [],
+          extraData: orderDetails.extraData ?? "0x",
+          offererBalancesAndApprovals: offerersBalancesAndApprovals[index],
+          offererOperator: allOffererOperators[index],
+        };
+        if (o.tips.length > 0) {
+          o.tips = adjustTipsForPartialFills(
+            o.tips,
+            o.unitsToFill || 1,
```
frankisawesome, line 1056: I don't 100% understand this code (the type is bigint | undefined), but I'm making the assumption here if unitsToFill is undefined it's because it's a ERC721 sell order where units to fill is 1. Tests seem to agree.

src/seaport.ts
```diff
@@ -1029,24 +1031,38 @@ export class Seaport {
...
-          orderDetails.tips?.map((tip) => ({
-            ...mapInputItemToOfferItem(tip),
-            recipient: tip.recipient,
-          })) ?? [],
-        extraData: orderDetails.extraData ?? "0x",
-        offererBalancesAndApprovals: offerersBalancesAndApprovals[index],
-        offererOperator: allOffererOperators[index],
-      }),
+      (orderDetails, index) => {
+        const o = {
```
ryanio, line 1035: could you use "order" instead of "o"?

frankisawesome: [931a81d](https://github.com/ProjectOpenSea/seaport-js/pull/559/commits/931a81d5e9bf4f8223efa891825fc60c236c3ac6)

ryanio (approved):

Issue #564, opened 2024-06-05
Incorrect tip amount scaling (only native currency) in partial fill scenario when using fulfillOrders


naveen-imtb:
### Component

API

### Have you ensured that all of these are up to date?

- [X] seaport-js
- [X] Node (minimum v16)

### What version of seaport-js are you on?

4.0.1

### What function is the bug in?

fulfillOrders

### Operating System

macOS (Apple Silicon)

### Describe the bug

Tip amounts in native currency are incorrectly scaled when performing a partial fill resulting in a much lower fee amount being charged than intended.

The most recent SDK version(4.0.1) includes a fix intended to address the issue where tip amounts were not at all scaled to the proportion of the order being filled in a partial fill scenario when using `fulfillOrders`. Unfortunately that fix has introduced a new bug where tip amounts are being incorrectly scaled (scaled to a much lower value than required) because the tip amounts sent to the Seaport contract are the scaled down values when they should infact reflect the original full order view.

Scenario:
* Create an order to sell `10` ERC1155 tokens at `1` ETH per token so total order value if `10` ETH.
* Include a 2% fee on the order which will be `0.2` ETH for the entire order or `0.02` ETH per token.
* Execute a partial fill for `2` units of the order by invoking `sdk.fulfillOrders()`

|                | Expected          | Actual                     | 
| -------------- | ----------------------- | -----------------------  |
| Cost of tokens | 2 ETH (2 * 1 ETH)       | 2 ETH (2 * 1 ETH)        |
| Fees           | 0.04 ETH (2 * 0.02 ETH) | 0.008 ETH (2 * 0.004 ETH)|
| Total          | 2.04 ETH                | 2.008 ETH                |

**Note:** This issue affects only partial fill scenarios using native currency which are executed by calling `fulfillOrders` and only seen in `v4.0.1` of the SDK.

naveen-imtb:
[PR](https://github.com/ProjectOpenSea/seaport-js/pull/565) to address this.

PR #565, opened 2024-06-05
bugfix: Fix native currency tip amount scaling logic for partial fills when using fulfillOrders

naveen-imtb:
## Motivation

Fixes [564](https://github.com/ProjectOpenSea/seaport-js/issues/564)

## Solution

The tip amount scaling logic has been moved into the `fulfillAvailableOrders` function where all the order's offer and consideration items are scaled proportional to the ratio of the order being filled.

This ensures that the `fulfillAvailableOrders` function has access to the original (unscaled) tip amounts to be passed to the Seaport contract as consideration items.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@opensea/seaport-js",
-  "version": "4.0.1",
+  "version": "4.0.2",
   "description": "[Seaport](https://github.com/ProjectOpenSea/seaport) is a new marketplace protocol for safely and efficiently buying and selling NFTs. This is a TypeScript library intended to make interfacing with the contract reasonable and easy.",
   "license": "MIT",
   "author": "OpenSea Developers",
--- a/src/seaport.ts
+++ b/src/seaport.ts
@@ -52,9 +52,8 @@ import {
   shouldUseBasicFulfill,
   validateAndSanitizeFromOrderStatus,
 } from "./utils/fulfill";
-import { getMaximumSizeForOrder, isCurrencyItem } from "./utils/item";
+import { isCurrencyItem } from "./utils/item";
 import {
-  adjustTipsForPartialFills,
   areAllCurrenciesSame,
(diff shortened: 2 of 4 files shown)


lfportal (approved):

ryanio:
thank you for the PR! could you add a test that helps validate that the math/scaling log is correct now? bonus points if it fails prior to the source change and passes after it :)

naveen-imtb:
> thank you for the PR! could you add a test that helps validate that the math/scaling log is correct now? bonus points if it fails prior to the source change and passes after it :)

Hi @ryanio, thank you for reviewing the PR.

Regarding your comment - my understanding is that the actual scaling happens in the smart contract and not in the SDK. In the case of partial fills, the on-chain smart contract uses the offer and consideration parameters that are submitted to the `fulfillAvailableAdvancedOrders()` in conjunction with the `numerator` and `denominator` values to determine the proposed fill ratio and scale down the consideration items and tips accordingly.

Therefore if incorrect order parameters, tips in the case of this specific bug, are passed to `fulfillOrders()` then the on-chain smart contract function `fulfillAvailableAdvancedOrders()` is called with those incorrect order parameters resulting in the scaled tip amounts to be inline with the incorrect inputs.

The expected assertion here would be for the unit test to verify that the correct order parameters (i.e the original form of the order) are passed to the `fulfillOrders()` SDK function to ensure that the correct values are used when invoking the on-chain smart contract function. 

I've [now extended](https://github.com/ProjectOpenSea/seaport-js/pull/565/commits/3962a762b3622081a1b2f39c6ca9a6e5a6fe7493) the test assertion to verify the function arguments.

As per your request, created another [PR](https://github.com/ProjectOpenSea/seaport-js/pull/572) to verify the presence of the bug, failing test assertion in the absence of the correct tip amounts scaling fix.

ryanio (approved):

PR #876, opened 2026-05-12
fix: support transaction overrides in fulfillOrders

aproskill81:
## Motivation

`fulfillOrders` did not expose transaction overrides even though `fulfillOrder` already supports them. This prevented callers from setting transaction-level fields such as `gasLimit`, `nonce`, or fee overrides when fulfilling multiple orders.

## Solution

Add `overrides` to the `fulfillOrders` input and pass it through to `fulfillAvailableOrders`.

The final transaction overrides preserve the SDK-calculated native token `value` while forwarding caller-provided fields. Added coverage that verifies a provided `gasLimit` is included in the built fulfillment transaction.

### DIFF
--- a/src/seaport.ts
+++ b/src/seaport.ts
@@ -937,6 +937,7 @@ export class Seaport {
    *                               Defaults to the zero address which means the offer goes to the fulfiller
    * @param input.domain optional domain to be hashed and appended to calldata
    * @param input.exactApproval optional boolean to indicate whether the approval should be exact or not
+   * @param input.overrides any transaction overrides the client wants, ignored if not set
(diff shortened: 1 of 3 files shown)


ryanio (approved):
LGTM. Cleanly closes the only gap in overrides plumbing across the public Seaport API; `{ ...overrides, value: totalNativeAmount }` matches the precedence used in `fulfillBasicOrder` / `fulfillStandardOrder`. Optional nit: consider harmonizing the three merge sites in `src/utils/fulfill.ts` (lines 291, 438, 720) on a single style — either reassign `overrides` in place or introduce a local `transactionOverrides` everywhere.

PR #911, opened 2026-07-23
fix: add missing await in getDomains to enable fallback on revert

### LINKED ISSUE #910: bug: getDomains fallback is unreachable due to missing await in try block
### Component

API, Utils

- [x] seaport-js
- [ ] Node (minimum v16)

4.1.4

getDomains

### Operating System

Linux

### Describe the bug

`getDomains()` includes a fallback for when the registry reverts due to too many domains. However, the `catch` block is dead code — it is never reached because the `try` block is missing `await`.

**Current code (`src/seaport.ts`):**

```typescript
public async getDomains(tag: string): Promise<string[]> {
  try {
    return this.domainRegistry.getDomains(tag)  // ← missing await
  } catch {
    // NEVER reached — async rejection bypasses this catch
    const totalDomains = await this.domainRegistry.getNumberOfDomains(tag)
    const domainArray = Promise.all(
      [...Array(Number(totalDomains)).keys()].map(i =>
        this.domainRegistry.getDomain(tag, i),
      ),
    )
    return await domainArray
  }
}
```

In an `async` function, `return promise` (without `await`) does not let the `catch` block intercept a rejection the rejection propagates directly to the caller.

**Proof:**

```typescript
async function broken() {
  try {
    return Promise.reject(new Error("revert"))  // no await
  } catch {
    console.log("fallback")  // never printed
  }
}

async function fixed() {
  try {
    return await Promise.reject(new Error("revert"))
  } catch {
    console.log("fallback")  // prints correctly
  }
}

broken()  // UnhandledPromiseRejection
fixed()   // "fallback"
```

**Impact:** When the registry reverts (too many domains), `getDomains()` throws an unhandled rejection instead of falling back to per-index queries. The fallback logic never runs.

**Fix:**

```typescript
public async getDomains(tag: string): Promise<string[]> {
  try {
    return await this.domainRegistry.getDomains(tag)  // add await
  } catch {
    const totalDomains = await this.domainRegistry.getNumberOfDomains(tag)
    return Promise.all(
      [...Array(Number(totalDomains)).keys()].map(i =>
        this.domainRegistry.getDomain(tag, i),
      ),
    )
  }
}
```

One line fix: add `await` before `this.domainRegistry.getDomains(tag)`.

Sertug17:
## Summary

Adds missing `await` in `getDomains()` so the fallback catch block can
actually be reached when the registry reverts.

## Problem

`return this.domainRegistry.getDomains(tag)` without `await` means any
async rejection bypasses the `catch` block entirely and propagates to the
caller as an unhandled rejection. The fallback logic (per-index query) was
dead code.

## Fix

```typescript
// before
return this.domainRegistry.getDomains(tag)

// after
return await this.domainRegistry.getDomains(tag)
```

## Related

Closes #910

### DIFF
--- a/src/seaport.ts
+++ b/src/seaport.ts
@@ -1219,7 +1219,7 @@ export class Seaport {
    */
   public async getDomains(tag: string): Promise<string[]> {
     try {
-      return this.domainRegistry.getDomains(tag)
+      return await this.domainRegistry.getDomains(tag)
     } catch {
       // If there are too many domains set under the tag, it will revert when trying to return in memory.
       // This fallback will manually query each index to get the full list of domains.


ryanio:
Thanks for the clear writeup and the nice repro in the issue — confirmed the missing `await` was indeed swallowing the async rejection before it could hit the `catch`. Fix looks correct. Kicking off CI review now before merging.

ryanio (approved):

Sertug17:
Thanks for the thorough review and quick turnaround, @ryanio! 🙏

PR #915, opened 2026-07-26
fix: use getItemToCriteriaMap in generateCriteriaResolvers

crazywriter1:
## Summary

- Fix criteria lookup in `generateCriteriaResolvers` to use `getItemToCriteriaMap` instead of indexing by item position in the offer/consideration array
- Align with the same pattern already used in `fulfill.ts`, `balanceAndApprovalCheck.ts`, and `item.ts`
- Throw a clear error when criteria input is missing for a criteria-based item

## Problem

`generateCriteriaResolvers` resolved criteria with:
    criterias[orderIndex][index]
Here, `index` is the item's position in the full offer/consideration array. However, `offerCriteria` / `considerationCriteria` are flat arrays with length equal to the number of criteria items only.
If a non-criteria item appears before a criteria item, lookup uses the wrong index. For example:
- offer = [ERC20, ERC721_WITH_CRITERIA]
- offerCriteria = [{ identifier, proof }]
The resolver looked up `criterias[0][1]` instead of `criterias[0][0]`, which could return `undefined` and crash at runtime.

## Solution

Use `getItemToCriteriaMap` to map criteria items to their corresponding `InputCriteria`, consistent with the rest of the codebase.

## Test plan

- [x] npm run lint
- [x] npm test

### DIFF
--- a/src/utils/criteria.ts
+++ b/src/utils/criteria.ts
@@ -47,7 +47,21 @@ export const generateCriteriaResolvers = ({
   ) =>
     criteriaItems.map(({ orderIndex, item, index, side }) => {
       const merkleRoot = item.identifierOrCriteria || "0"
-      const inputCriteria: InputCriteria = criterias[orderIndex][index]
+      const items =
+        side === Side.OFFER
+          ? orders[orderIndex].parameters.offer
+          : orders[orderIndex].parameters.consideration
+      const inputCriteria = getItemToCriteriaMap(
+        items,
+        criterias[orderIndex] ?? [],
+      ).get(item)
+
+      if (!inputCriteria) {
+        throw new Error(
+          "You must supply the appropriate criterias for criteria based items",
+        )
+      }
+
       return {
         orderIndex,
         index,
--- a/test/utils/criteria.spec.ts
+++ b/test/utils/criteria.spec.ts
@@ -1,0 +1,181 @@
+import { expect } from "chai"
+import { ZeroAddress, ZeroHash } from "ethers"
+import { ItemType, OrderType, Side } from "../../src/constants"
+import type {
+  ConsiderationItem,
+  InputCriteria,
+  OfferItem,
+  Order,
+} from "../../src/types"
+import { generateCriteriaResolvers } from "../../src/utils/criteria"
+
+const offerItem = (
+  itemType: ItemType,
+  identifierOrCriteria = "0",
+): OfferItem => ({
+  itemType,
+  token: ZeroAddress,
+  identifierOrCriteria,
+  startAmount: "1",
+  endAmount: "1",
+})
+
+const considerationItem = (
+  itemType: ItemType,
+  identifierOrCriteria = "0",
+): ConsiderationItem => ({
+  ...offerItem(itemType, identifierOrCriteria),
+  recipient: ZeroAddress,
+})
+
+const order = (
+  offer: OfferItem[],
+  consideration: ConsiderationItem[] = [],
+): Order => ({
+  parameters: {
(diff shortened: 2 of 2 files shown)


ryanio (commented):
Confirmed this is a real bug and the fix is the right one. CI is green.

Reproduced against `main` — the trigger is any order where a non-criteria item precedes a criteria item, because `index` is the position in the *full* offer array while `offerCriterias[orderIndex]` is a flat array holding only the criteria-based entries:

```ts
offer: [
  { itemType: ERC20 },                  // index 0, not criteria
  { itemType: ERC721_WITH_CRITERIA },   // index 1, IS criteria
]
offerCriterias: [[{ identifier: "42", proof: [] }]]   // length 1
```

`criterias[0][1]` is `undefined`, so `inputCriteria.identifier` throws:

```
TypeError: Cannot read properties of undefined (reading 'identifier')
```

With this patch applied the same input resolves correctly:

```
[{ orderIndex: 0, index: 1, side: 0, identifier: "42", criteriaProof: [] }]
```

Routing through `getItemToCriteriaMap` is right, and it matches how `fulfill.ts`, `balanceAndApprovalCheck.ts` and `item.ts` already pair items with criteria. The `Map` is keyed by item identity and you re-derive `items` from `orders[orderIndex].parameters`, which is the same array `criteriaItems` was built from, so the `.get(item)` lookups hit. Note the emitted `index` is still the full-array position, which is what Seaport's criteria resolvers expect — the fix changes only which criteria is looked up, not the reported index.

Worth noting in the description that `criterias[orderIndex] ?? []` fixes a second crash on the same line: with the `[[]]` defaults and more than one order, `criterias[1]` is `undefined` and the old `[index]` access threw before any criteria lookup happened.

### One thing needed: a regression test

CI reports 122 passing, identical to `main`, so nothing here exercises the fix. Given this is a correctness bug in criteria resolution that was silently throwing a `TypeError`, it should come with a test that fails without the patch — otherwise the next refactor can reintroduce it. The two-item offer above is enough, and `test/utils/criteria.spec.ts` would be a natural home. A case with two orders using the `[[]]` default would cover the `?? []` path too.

Happy to push that test to your branch if you'd prefer — say the word and I'll add it, otherwise this is ready to merge as soon as it has coverage.

### Minor

`getItemToCriteriaMap(items, criterias[orderIndex] ?? [])` is rebuilt on every iteration of the `.map`, so an order with *n* criteria items walks its item array *n* times and allocates *n* maps. Item counts are small so this is not a practical concern, but hoisting one map per `(orderIndex, side)` outside the callback would read more cleanly. Entirely optional.

ryanio:
Pushed a regression test to your branch (5d53ca2) so this can go into the 4.1.6 release — thanks for catching this, the diagnosis in the description was exactly right.

`test/utils/criteria.spec.ts` covers the bug you found plus the adjacent paths: multiple criteria items in one order, offer and consideration resolved from their own flat arrays, multiple orders, and the missing-criteria error. Verified it fails 5 tests against the unpatched `criteria.ts` and passes 128 with your fix, so it genuinely locks the behavior in.

I left the `getItemToCriteriaMap` call inside the `.map` as you wrote it — the hoisting note was optional and not worth churning the diff for.

Your commits are untouched at the base of the branch, so the fix stays yours.

ryanio (approved):
Approving. `tests` and `lint` green — 129 passing with the regression suite added in 5d53ca2, verified to fail 5 tests without the fix.

Thanks for finding this one; the flat-array-vs-item-position mismatch was a subtle bug. Going into 4.1.6.

PR #917, opened 2026-07-26
fix: dedupe approval actions by token and operator

### LINKED ISSUE #916: Approval deduplication ignores operator when conduit differs for the same token
### Component

API, Utils

- [x] seaport-js
- [ ] Node (minimum v16)

4.1.5

createBulkOrders, getApprovalActions

### Operating System

Linux

### Describe the bug

Approval actions are deduplicated by token address only, ignoring operator (Seaport contract vs conduit). When the same token needs approvals to different operators, one approval is silently dropped.

This affects:
- getApprovalActions in src/utils/approval.ts
- createBulkOrders approval merging in src/seaport.ts

### getApprovalActions (approval.ts)

Insufficient approvals are filtered with logic that only compares token addresses between consecutive entries. If the same token requires approvals to different operators (e.g. Seaport vs OpenSea conduit), only one approval action survives.

### createBulkOrders (seaport.ts)

Bulk order creation merges per-order approval actions by checking only:

    allApprovalActions.find(a => a.token === approval.token)

A second approval for the same token but a different operator is dropped.

## Example

Bulk-creating orders that offer the same ERC20/NFT with mixed conduit keys:

- Order 1: conduitKey = NO_CONDUIT → approve Seaport
- Order 2: conduitKey = OPENSEA_CONDUIT_KEY → approve OpenSea conduit

Expected: 2 approval actions (one per operator)
Actual: 1 approval action; the other operator never gets approved

Fulfillment can then fail on-chain due to missing approval.

## Why this matters

operator is resolved from conduitKeyToConduit[conduitKey] per order. Different conduitKey values can require different operators for the same token. Token-only dedup is incorrect in that case.

Current bulk order tests only cover a single conduit (NO_CONDUIT), so this path is untested.

## Proposed fix

Deduplicate approvals by (token, operator) instead of token alone. For exact ERC721 approvals, also include identifierOrCriteria in the dedup key.

## Affected code

- src/utils/approval.ts — getApprovalActions
- src/seaport.ts — createBulkOrders approval merge loop (~lines 258-264)

crazywriter1:
Fixes #916 

## Summary

- Deduplicate approval actions by (token, operator) instead of token alone
- Export `getApprovalDedupKey` and reuse it in `createBulkOrders`
- For exact ERC721 approvals, include `identifierOrCriteria` in the dedup key

## Problem

Approval actions were deduplicated by token address only. When the same token required approvals to different operators (e.g. Seaport vs OpenSea conduit), one approval was silently dropped in:
- `getApprovalActions` (`approval.ts`)
- `createBulkOrders` approval merge (`seaport.ts`)

## Test plan

- [x] npm run lint
- [x] npm test -- --grep "create bulk orders"

### DIFF
--- a/src/seaport.ts
+++ b/src/seaport.ts
@@ -45,7 +45,7 @@ import type {
   SeaportContract,
   TipInputItem,
 } from "./types"
-import { getApprovalActions } from "./utils/approval"
+import { getApprovalActions, getApprovalDedupKey } from "./utils/approval"
 import {
   getBalancesAndApprovals,
   validateOfferBalancesAndApprovals,
@@ -255,11 +255,18 @@ export class Seaport {
 
       allOrderComponents.push(orderComponents)
 
-      // Dedupe approvals by token address
+      // Dedupe approvals by token and operator (and identifier for exact ERC721)
+      const seenApprovalKeys = new Set(
+        allApprovalActions.map(approval =>
+          getApprovalDedupKey(approval, Boolean(exactApproval)),
+        ),
+      )
+
       for (const approval of approvalActions) {
-        if (
-          allApprovalActions.find(a => a.token === approval.token) === undefined
-        ) {
+        const key = getApprovalDedupKey(approval, Boolean(exactApproval))
+
+        if (!seenApprovalKeys.has(key)) {
+          seenApprovalKeys.add(key)
           allApprovalActions.push(approval)
         }
       }
--- a/src/utils/approval.ts
+++ b/src/utils/approval.ts
@@ -32,6 +32,29 @@ export const approvedItemAmount = async (
   return MAX_INT
 }
 
+export const getApprovalDedupKey = (
+  approval: {
+    token: string
+    operator: string
+    itemType: ItemType
+    identifierOrCriteria: string
+  },
+  exactApproval: boolean,
+): string => {
+  const token = approval.token.toLowerCase()
+  const operator = approval.operator.toLowerCase()
+
+  if (
+    exactApproval &&
+    isErc721Item(approval.itemType) &&
+    !isErc1155Item(approval.itemType)
+  ) {
+    return `${token}:${operator}:${approval.identifierOrCriteria}`
+  }
+
+  return `${token}:${operator}`
+}
+
 /**
  * Get approval actions given a list of insufficient approvals.
  */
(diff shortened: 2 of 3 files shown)


ryanio (approved):
Confirmed the bug and the fix. Approving — CI is green (123 passing, including your new case).

The diagnosis in #916 is right: `operator` is derived per-order from `conduitKeyToConduit[conduitKey]`, so two orders offering the same token under different conduit keys legitimately need two approvals, and both dedup sites collapsed them to one. Verified the supporting details — `ApprovalAction` carries `operator` and `identifierOrCriteria`, `exactApproval` is in scope at the `createBulkOrders` call site, and `InsufficientApprovals` matches the shape your test constructs.

Including `identifierOrCriteria` only under `exactApproval` for ERC721 is the correct condition, since that is exactly the branch where `getApprovalActions` emits per-token-id `approve(operator, identifierOrCriteria)` rather than `setApprovalForAll`.

Three notes, none blocking.

**The `getApprovalActions` rewrite fixes a second latent bug worth mentioning in the description.** The old filter only compared *adjacent* entries, so non-consecutive duplicates both survived. Given `[A(token1), B(token2), C(token1)]` with one operator, `main` emits three approval actions — two redundant `setApprovalForAll` calls for `token1`. Your version emits `[B, C]`. That is a real improvement beyond the stated scope, and it makes the change slightly broader than "dedupe by token and operator," so it's worth calling out.

The reverse/dedupe/reverse pattern also preserves both the original relative order of survivors and the old keep-the-last-occurrence semantics, which is why the existing suite is unaffected.

**`!isErc1155Item(...)` is redundant.** The two predicates are disjoint over their value sets:

```ts
export const isErc721Item = (itemType) => [ItemType.ERC721, ItemType.ERC721_WITH_CRITERIA].includes(itemType)
export const isErc1155Item = (itemType) => [ItemType.ERC1155, ItemType.ERC1155_WITH_CRITERIA].includes(itemType)
```

`isErc721Item(x) && !isErc1155Item(x)` is just `isErc721Item(x)`. Harmless, but it reads as though the cases can overlap. (The `!isErc1155` guard further down in `getApprovalActions` *is* load-bearing, because that branch is entered for either family — so dropping only the one in `getApprovalDedupKey` is the right call.)

**The test doesn't cover the `createBulkOrders` path this PR also changed.** It calls `getApprovalActions` directly, so the approval-merge loop in `seaport.ts` — the half #916 flagged at lines 258-264 — is still untested. Since it's a unit test of an `src/utils/approval.ts` export, `test/utils/approval.spec.ts` is also a more natural home than `create-bulk-orders.spec.ts`. A bulk-order case with two orders that mix `NO_CONDUIT` and `OPENSEA_CONDUIT_KEY` on the same token, asserting two approval actions, would close the gap the issue explicitly identified as untested. Follow-up is fine.

PR #923, opened 2026-07-30
fix: guard against division by zero when startTime equals endTime in getPresentItemAmount

### LINKED ISSUE #922: bug: RangeError division by zero in getPresentItemAmount when startTime equals endTime
## Summary

`getPresentItemAmount` in `src/utils/item.ts` throws an unhandled `RangeError: Division by zero` when an order has `startTime === endTime`.

## Root Cause

```typescript
const duration = endTimeBn - startTimeBn  // 0n when startTime === endTime

// no guard exists, so this crashes:
return (
  (startAmountBn * remaining + endAmountBn * elapsed + ...) /
  duration  // RangeError: Division by zero
)
```

BigInt division by zero throws `RangeError` in JavaScript unlike regular numbers which return `Infinity`.

## Reproduction

```js
node -e "
const startAmountBn = 1000n, endAmountBn = 2000n
const startTimeBn = 1000n, endTimeBn = 1000n
const duration = endTimeBn - startTimeBn
const elapsed = 0n, remaining = 0n
const result = (startAmountBn * remaining + endAmountBn * elapsed) / duration
"
// RangeError: Division by zero
```

## Impact

Any fulfillment attempt on an order where `startTime === endTime` crashes. `fulfillBasicOrder`, `fulfillStandardOrder`, and `fulfillAvailableOrders` all pass order `startTime`/`endTime` directly into `timeBasedItemParams`, which flows into `getPresentItemAmount`.

## Expected Behavior

The Seaport contract's `AmountDeriver` handles zero-duration orders by returning `endAmount`. The JS SDK should match this behavior.

## Fix

```typescript
const duration = endTimeBn - startTimeBn
if (duration === 0n) {
  return endAmountBn
}
```

Sertug17:
## Summary

Adds a zero-duration guard in `getPresentItemAmount` to prevent `RangeError: Division by zero` when an order's `startTime === endTime`.

## Problem

When `startTime === endTime`, `duration = 0n`. BigInt division by `0n` throws `RangeError` in JavaScript. There was no guard, so any fulfillment call on such an order crashed in `fulfillBasicOrder`, `fulfillStandardOrder`, and `fulfillAvailableOrders`.

## Fix

```typescript
const duration = endTimeBn - startTimeBn
if (duration === 0n) {
  return endAmountBn  // matches Seaport contract AmountDeriver behavior
}
```

## Testing

**Before fix:**

node -e "... / 0n"
// RangeError: Division by zero

**After fix 3 new unit tests added (`test/utils/item.spec.ts`):**

getPresentItemAmount
zero-duration order (startTime === endTime)
✔ returns endAmount without throwing RangeError for ascending amounts
✔ returns endAmount without throwing RangeError for descending amounts
✔ returns endAmount for consideration items
3 passing (3ms)

All 118 existing tests continue to pass.

Closes #922

### DIFF
--- a/src/utils/item.ts
+++ b/src/utils/item.ts
@@ -55,6 +55,9 @@ export const getPresentItemAmount = ({
   const endTimeBn = BigInt(endTime)
 
   const duration = endTimeBn - startTimeBn
+  if (duration === 0n) {
+    return endAmountBn
+  }
   const isAscending = endAmountBn > startAmountBn
   const adjustedBlockTimestamp = BigInt(
     isAscending
--- a/src/utils/item.ts
+++ b/src/utils/item.ts
@@ -55,9 +55,6 @@ export const getPresentItemAmount = ({
   const endTimeBn = BigInt(endTime)
 
   const duration = endTimeBn - startTimeBn
-  if (duration === 0n) {
-    return endAmountBn // zero-duration: treat order as at its end
-  }
   const isAscending = endAmountBn > startAmountBn
   const adjustedBlockTimestamp = BigInt(
     isAscending
@@ -67,6 +64,13 @@ export const getPresentItemAmount = ({
 
   if (adjustedBlockTimestamp < startTimeBn) {
     return startAmountBn
+  }
+
+  // A zero-duration order would divide by zero below. Seaport rejects these at
+  // _verifyTime (endTime <= block.timestamp reverts InvalidTime), so the value
+  // is inert; returning endAmount just avoids a RangeError during derivation.
+  if (duration === 0n) {
+    return endAmountBn
   }
 
   const elapsed =
--- a/src/utils/item.ts
+++ b/src/utils/item.ts
@@ -56,7 +56,7 @@ export const getPresentItemAmount = ({
 
   const duration = endTimeBn - startTimeBn
   if (duration === 0n) {
-    return endAmountBn
+    return endAmountBn // zero-duration: treat order as at its end
   }
   const isAscending = endAmountBn > startAmountBn
   const adjustedBlockTimestamp = BigInt(
--- a/test/utils/item.spec.ts
+++ b/test/utils/item.spec.ts
@@ -1,0 +1,51 @@
+import { expect } from "chai"
+import { getPresentItemAmount } from "../../src/utils/item"
+
+describe("getPresentItemAmount", () => {
+  describe("zero-duration order (startTime === endTime)", () => {
+    it("returns endAmount without throwing RangeError for ascending amounts", () => {
+      const result = getPresentItemAmount({
+        startAmount: "1000",
+        endAmount: "2000",
+        timeBasedItemParams: {
+          startTime: "1000",
+          endTime: "1000",
+          currentBlockTimestamp: 1001,
+          ascendingAmountTimestampBuffer: 0,
+          isConsiderationItem: false,
+        },
+      })
+      expect(result).to.equal(2000n)
+    })
+
+    it("returns endAmount without throwing RangeError for descending amounts", () => {
+      const result = getPresentItemAmount({
+        startAmount: "2000",
+        endAmount: "1000",
+        timeBasedItemParams: {
+          startTime: "1000",
+          endTime: "1000",
+          currentBlockTimestamp: 1001,
+          ascendingAmountTimestampBuffer: 0,
+          isConsiderationItem: false,
+        },
+      })
+      expect(result).to.equal(1000n)
+    })
+
+    it("returns endAmount for consideration items", () => {
+      const result = getPresentItemAmount({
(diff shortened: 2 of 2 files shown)


ryanio (commented):
Thanks for filing #922 with a clean reproduction and turning it around the same day. The crash is real and worth fixing — CI is now green here (125 passing, so your three new `test/utils/item.spec.ts` cases are being picked up by the runner).

Two things before this goes in.

### 1. The guard is placed one step too early

Putting the `duration === 0n` check above the `adjustedBlockTimestamp < startTimeBn` branch changes behavior for a case that never crashed. A zero-duration order whose start time is still in the future already returned `startAmount` via the not-yet-started branch:

| case | `main` | this PR | guard moved below the not-started check |
|---|---|---|---|
| zero-duration, already started (your test 1) | `RangeError` | `2000` | `2000` |
| zero-duration, descending (your test 2) | `RangeError` | `1000` | `1000` |
| **zero-duration, not yet started** | **`1000`** | **`2000`** | **`1000`** |
| normal ascending order | `1500` | `1500` | `1500` |

Only the third row differs, and it is not one of the crashing cases. `getPresentItemAmount` is exported, so that is a behavior change in published API surface for no benefit.

Moving the guard four lines down fixes the `RangeError` and leaves that row alone:

```ts
  const duration = endTimeBn - startTimeBn
  const isAscending = endAmountBn > startAmountBn
  const adjustedBlockTimestamp = BigInt(
    isAscending
      ? currentBlockTimestamp + ascendingAmountTimestampBuffer
      : currentBlockTimestamp,
  )

  if (adjustedBlockTimestamp < startTimeBn) {
    return startAmountBn
  }

  if (duration === 0n) {
    return endAmountBn
  }
```

All three of your existing tests still pass with it there, since they all use `currentBlockTimestamp > startTime`. A fourth case pinning the not-yet-started path (`startTime === endTime === 2000`, `currentBlockTimestamp: 1000`, expecting `startAmount`) would lock the distinction in.

### 2. The rationale in the description isn't quite right

Both the PR and #922 say returning `endAmount` "matches Seaport contract AmountDeriver behavior." That early return is gated on `startAmount != endAmount`, not on zero duration:

```solidity
// ReferenceAmountDeriver.sol
if (startAmount != endAmount) { /* ... divides by duration ... */ }
return endAmount;  // reached only when startAmount == endAmount
```

And a zero-duration order is never fulfillable at all, because `_verifyTime` rejects `endTime <= block.timestamp`:

```solidity
if (startTime > block.timestamp || endTime <= block.timestamp) { revert InvalidTime(startTime, endTime); }
```

So `startTime === endTime` always reverts `InvalidTime` onchain — the contract never reaches the division, which is why it has no zero-duration guard to match.

That doesn't undermine the fix; it just changes the justification. The value here is that the SDK should not blow up with an opaque `RangeError` while deriving an amount. Returning `endAmount` lets the caller proceed to a clear Seaport `InvalidTime` revert instead. Could you reword to something like "avoid a `RangeError` on a degenerate order; the value is inert because Seaport rejects zero-duration orders at `_verifyTime`"? Worth keeping accurate for whoever reads this next.

Happy to push the guard move to your branch if you'd rather not round-trip.

ryanio:
Pushed the guard move to your branch (0bf8237) so this can go into the 4.1.6 release — thanks for the quick turnaround on this, and for filing #922 with a reproduction that made it easy to confirm.

Two changes on top of your commits:

- Moved the `duration === 0n` check below the `adjustedBlockTimestamp < startTimeBn` branch. Your three tests all pass unchanged, since each uses `currentBlockTimestamp > startTime`.
- Added a fourth case pinning the not-yet-started path, and reworded the comment to say why the contract has no equivalent guard rather than implying it does.

Your commits are untouched at the base of the branch, so the fix stays yours.

ryanio (approved):
Approving.  and Protected by Socket Firewall

=== Socket Firewall === green — 127 passing, including the not-yet-started case added in 0bf8237. Guard now sits below the not-yet-started branch, so the `RangeError` is fixed with no change to any path that previously worked.

Thanks again for the fast, well-documented fix. Going into 4.1.6.

ryanio:
(My approval note above got mangled by a shell quoting issue — the intended text:)

Approving. `tests`, `lint` and `coverage` are all green — 127 passing, including the not-yet-started case added in 0bf8237. The guard now sits below the not-yet-started branch, so the `RangeError` is fixed with no change to any path that previously worked.

Thanks again for the fast, well-documented fix. Going into 4.1.6.

Sertug17:
Thanks for the thorough review and the correction on the contract behavior @ryanio! Good to know the onchain path reverts with InvalidTime rather than reaching the division the SDK guard still makes sense to surface a clean revert instead of crashing during amount derivation. Appreciate the quick turnaround on v4.1.6!

PR #927, opened 2026-07-31
fix: include orderIndex in ERC721 de-dup key to prevent cross order aggregation

### LINKED ISSUE #926: ERC721 items from different orders incorrectly aggregated in fulfillAvailableAdvancedOrders
### Component

Utils

- [x] seaport-js
- [ ] Node (minimum v16)

latest

generateFulfillOrdersFulfillments

### Operating System

Linux

### Describe the bug

`generateFulfillOrdersFulfillments` in `src/utils/fulfill.ts` generates incorrect
fulfillment components when two orders from the same offerer both have an ERC721 item
at the same array position (`offer[0]`) with the same token address and token identifier.

The de-duplication key uses only `itemIndex`, not `orderIndex`, so the two items collide
and get merged into a single fulfillment component. Seaport then tries to transfer the
ERC721 with `amount = 2`, which always reverts with `InvalidERC721TransferAmount`.

The code comment directly above the affected lines already states the intent:
> "We tack on the index to ensure that erc721s can never be aggregated and instead must be in separate arrays"

But `itemIndex` alone fails for cross-order cases.

## Affected lines

`src/utils/fulfill.ts` lines 840 and 864:

```ts
// offer loop (line 840) BUGGY
})}${isErc721Item(item.itemType) ? itemIndex : ""}`

// consideration loop (line 864) BUGGY
})}${isErc721Item(item.itemType) ? itemIndex : ""}`
```

## Reproduction

1. Same offerer has two active listings for the same ERC721 (token `0xNFT`, id `1`)
2. Call `fulfillAvailableAdvancedOrders` with both orders
3. SDK generates: `offerFulfillments = [[{orderIndex:0,itemIndex:0},{orderIndex:1,itemIndex:0}]]`
4. Seaport aggregates → `transferFrom(offerer, recipient, tokenId)` with `amount=2`
5. Transaction reverts: `InvalidERC721TransferAmount`

**Expected:** order 0 fills, order 1 is skipped as unavailable.  
**Actual:** entire transaction reverts.

## Fix

Change `itemIndex` → `` `${orderIndex}-${itemIndex}` `` in both loops:

```ts
// offer loop
})}${isErc721Item(item.itemType) ? `${orderIndex}-${itemIndex}` : ""}`

// consideration loop
})}${isErc721Item(item.itemType) ? `${orderIndex}-${itemIndex}` : ""}`
```

Also, the return type annotation of `generateFulfillOrdersFulfillments` is incorrect:
`FulfillmentComponentStruct[]` should be `FulfillmentComponentStruct[][]`
(currently masked by `as any` at the call site).

Sertug17:
## Summary

Fixes #926

`generateFulfillOrdersFulfillments` in `src/utils/fulfill.ts` built the ERC721
de-duplication key using only `itemIndex`, not `orderIndex`. When two orders from
the same offerer both have an ERC721 at `offer[0]` with the same token/identifier,
the keys collide and the items are merged into a single fulfillment component.
Seaport then attempts `amount = 2` on the ERC721 transfer and always reverts with
`InvalidERC721TransferAmount`.

The existing code comment already states the intent:
> "We tack on the index to ensure that erc721s can never be aggregated"

But `itemIndex` alone doesn't achieve this across orders.

## Changes

- `src/utils/fulfill.ts`: Change `itemIndex` → `` `${orderIndex}-${itemIndex}` `` in both the offer and consideration aggregation loops
- `src/utils/fulfill.ts`: Correct return type from `FulfillmentComponentStruct[]` to `FulfillmentComponentStruct[][]` (was masked by `as any` at call site)

## Testing

All 118 existing tests pass.

### DIFF
--- a/src/utils/fulfill.ts
+++ b/src/utils/fulfill.ts
@@ -774,8 +774,8 @@ export function fulfillAvailableOrders({
               ),
             })
           : [],
-        offerFulfillments as any,
-        considerationFulfillments as any,
+        offerFulfillments,
+        considerationFulfillments,
         conduitKey,
         recipientAddress,
         advancedOrdersWithTips.length,
@@ -814,12 +814,12 @@ export function generateFulfillOrdersFulfillments(
 
   const offerAggregatedFulfillments: Record<
     string,
-    FulfillmentComponentStruct
+    FulfillmentComponentStruct[]
   > = {}
 
   const considerationAggregatedFulfillments: Record<
     string,
-    FulfillmentComponentStruct
+    FulfillmentComponentStruct[]
   > = {}
 
   ordersMetadata.forEach(
@@ -840,9 +840,9 @@ export function generateFulfillOrdersFulfillments(
         })}${isErc721Item(item.itemType) ? `${orderIndex}-${itemIndex}` : ""}`
 
         offerAggregatedFulfillments[aggregateKey] = [
-          ...((offerAggregatedFulfillments[aggregateKey] ?? []) as any),
+          ...(offerAggregatedFulfillments[aggregateKey] ?? []),
           { orderIndex, itemIndex },
-        ] as any
+        ]
       })
     },
   )
@@ -864,10 +864,9 @@ export function generateFulfillOrdersFulfillments(
           })}${isErc721Item(item.itemType) ? `${orderIndex}-${itemIndex}` : ""}`
 
           considerationAggregatedFulfillments[aggregateKey] = [
-            ...((considerationAggregatedFulfillments[aggregateKey] ??
-              []) as any),
+            ...(considerationAggregatedFulfillments[aggregateKey] ?? []),
             { orderIndex, itemIndex },
-          ] as any
+          ]
         },
       )
     },
--- a/src/utils/fulfill.ts
+++ b/src/utils/fulfill.ts
@@ -797,8 +797,8 @@ export function fulfillAvailableOrders({
 export function generateFulfillOrdersFulfillments(
   ordersMetadata: FulfillOrdersMetadata,
 ): {
-  offerFulfillments: FulfillmentComponentStruct[]
-  considerationFulfillments: FulfillmentComponentStruct[]
+  offerFulfillments: FulfillmentComponentStruct[][]
+  considerationFulfillments: FulfillmentComponentStruct[][]
 } {
   const hashAggregateKey = ({
     sourceOrDestination,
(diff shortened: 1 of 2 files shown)


Sertug17:
Added a unit test in `test/utils/fulfill.spec.ts` that reproduces the cross-order ERC721 collision and verifies the fix. Also confirms ERC1155s with the same token/id are still correctly aggregated across orders.

Sertug17:
@ryanio could you approve the workflow run so CI can execute?

ryanio:
yes, thanks for the PR!

Sertug17:
Updated: also fixed the internal accumulator types (`Record<string, FulfillmentComponentStruct[]>`) and removed all `as any` casts from the function build and 120 tests pass locally.

ryanio:
Thanks for tracking this down, the diagnosis is right and matches what the comment above those lines already claimed. I reviewed it and picked it up in #932, which keeps your two commits so the authorship stays with you.

Two things needed sorting before it could go in. The return type correction to `FulfillmentComponentStruct[][]` left the two accumulators declared as `Record<string, FulfillmentComponentStruct>`, so `Object.values()` still yields `FulfillmentComponentStruct[]` and `tsc` fails:

```
src/utils/fulfill.ts(877,5): error TS2322: Type 'FulfillmentComponentStruct[]' is not assignable to type 'FulfillmentComponentStruct[][]'.
```

That is why all three jobs went red at the build step, before any test ran, so the "all 118 existing tests pass" note in the description could not have been from this branch. #932 types the records as arrays, which also lets the four `as any` casts go away.

The branch was also 23 commits behind `main` (including #915, which touches this file), so #932 rebases it. Beyond that: the spec moved out of `test/utils/` since that directory holds shared helpers while every other spec sits at `test/` top level, the fixture now uses the real `BalancesAndApprovals` shape rather than `{ balances: [], approvals: [] } as any`, and there is coverage for the consideration loop your change also touches.

One note on impact. The key only collides when two orders share offerer (or recipient), token, and identifier, which means the same physical NFT, and both orders can never settle. So the batch still reverts after the fix, just at the token's `transferFrom` instead of with `InvalidERC721TransferAmount`. It is worth doing because it makes the code match its documented invariant, but it does not turn a failing fulfillment into a succeeding one.

Sertug17:
@ryanio should I close #927 in favour of #932, or will you merge this one?

ryanio:
Correction on my last comment: you had already pushed 9870800 fixing the accumulator types and dropping the `as any` casts about two minutes before my comment landed, and I was reading the older head. Your fix is byte-for-byte the same as the one I wrote in #932. Apologies for describing it as still outstanding.

#932 is merged, so this is now in `main` and I'm closing this one. Your two original commits went in with your authorship intact, and the third is credited via `Co-authored-by`.

What #932 carried on top of your final state, for the record:

- Rebased onto `main`. This branch was 23 commits behind, including #915 which touches this same file.
- `biome check` was failing on the test file, and CI never got far enough to tell you because the build step ran first.
- Moved the spec out of `test/utils/`. That directory holds shared helpers; every other spec sits at `test/` top level.
- `offererBalancesAndApprovals` in the fixture was `{ balances: [], approvals: [] } as any`, but `BalancesAndApprovals` is a flat array. The `as any` was hiding the mismatch. Also swapped the placeholder token for a real hex address.
- Added coverage for the consideration loop and the within-one-order case, since the key change affects both.

Thanks for the careful writeup on #926, and for chasing the type error down yourself before I got to it.

PR #932, opened 2026-08-07
fix: include orderIndex in the ERC721 fulfillment de-dup key

### LINKED ISSUE #926: ERC721 items from different orders incorrectly aggregated in fulfillAvailableAdvancedOrders

ryanio:
Builds on @Sertug17's #927 (their commits are preserved here). Closes #926, supersedes #927.

## The fix

`generateFulfillOrdersFulfillments` builds a de-duplication key per item and groups items that share a key into one fulfillment component array. For ERC721 the key gets a suffix so those items can never be aggregated, and the comment above it says exactly that:

> We tack on the index to ensure that erc721s can never be aggregated and instead must be in separate arrays

The suffix was `itemIndex` alone, which only separates ERC721s within a single order. Two orders from the same offerer whose ERC721 at the same offer index has the same token and identifier produce the same key and collapse into one group, and Seaport then derives a single transfer with `amount = 2`. Same shape on the consideration side, keyed on recipient instead of offerer. Using `${orderIndex}-${itemIndex}` makes the suffix unique per item across the whole batch, which is what the comment describes.

## What this adds on top of #927

#927 also corrected the return type from `FulfillmentComponentStruct[]` to `FulfillmentComponentStruct[][]`, but left the two accumulators declared as `Record<string, FulfillmentComponentStruct>`. `Object.values()` on those yields `FulfillmentComponentStruct[]`, so `tsc` reported TS2322 on both returned fields and every CI job failed at the build step before any test ran:

```
src/utils/fulfill.ts(877,5): error TS2322: Type 'FulfillmentComponentStruct[]' is not assignable to type 'FulfillmentComponentStruct[][]'.
```

This branch declares the records as `Record<string, FulfillmentComponentStruct[]>`, which is what they have always held at runtime. That makes the four `as any` casts unnecessary, so they are gone and the two `fulfillAvailableAdvancedOrders` arguments are now checked against the contract ABI instead of being cast away.

It also rebases onto current `main` (#927 was 23 commits behind, including #915 which touches this file), moves the spec out of `test/utils/` since that directory holds shared helpers while every other spec sits at `test/` top level, gives the fixture the real `BalancesAndApprovals` shape instead of `{ balances: [], approvals: [] } as any`, uses valid hex addresses, and covers the consideration loop and the same-order case that the key change also affects.

## Scope

Splitting a group is always safe. Seaport skips zero-amount executions, and the comment in `OrderCombiner._performFinalChecksAndExecuteOrders` names this exact case:

> the corresponding fulfillment contained only items on orders that are unavailable or are out of range

So a group holding only components from unavailable orders is already normal today and needs no new handling.

Worth setting expectations on the payoff, though: the key collides only when two orders share offerer (or recipient), token, and identifier, which means the same physical NFT. Both orders can never actually settle, so the batch still reverts after this change, just at the token's `transferFrom` rather than with `InvalidERC721TransferAmount`. Treat this as making the code do what its comment claims rather than as unblocking a fulfillment that would otherwise succeed.

## Testing

`npm run build`, `biome check .` and `npx hardhat test` all pass locally: 138 passing, up from 133 on `main`. Reverting just the key change makes the two cross-order tests fail and the rest pass, so they guard the behavior rather than restating it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Sertug17:
Thanks @ryanio! Glad it made it in and appreciate the detailed explanation on the type fix. Looking forward to contributing more.

PR #935, opened 2026-08-08
fix: use getApprovalDedupKey in addApprovalIfNeeded for exactApproval correctness

### LINKED ISSUE #934: fulfillOrders with exactApproval: true silently drops approval actions for ERC721 tokens from the same collection
### Component

Utils

- [x] seaport-js
- [ ] Node (minimum v16)

4.1.7

fulfillAvailableOrders / addApprovalIfNeeded

### Operating System

None

## Summary

When `fulfillOrders` is called with `exactApproval: true` and multiple orders whose consideration includes ERC721 tokens from the **same contract** (different token IDs), approval actions for every NFT after the first are silently dropped. The fulfiller's on-chain transaction reverts because those tokens were never approved.

## Root cause

In `fulfillAvailableOrders` (`src/utils/fulfill.ts`), the inner `addApprovalIfNeeded` closure deduplicates insufficient approvals by **token address only**:

```ts
// current (buggy)
!totalInsufficientApprovals.find(
  approval => approval.token === insufficientApproval.token,
)
```

For `exactApproval: false` this is harmless a single `setApprovalForAll` covers every token ID from the same contract. But for `exactApproval: true`, the library must emit a separate `approve(tokenId)` call per NFT (via `getApprovalDedupKey` returning `token:operator:identifierOrCriteria`). The token-only dedup silently discards every additional NFT from the same collection before `getApprovalActions` ever sees it.

The sibling function `createBulkOrders` already uses `getApprovalDedupKey` correctly for this same purpose — the two code paths are inconsistent.

## Reproduction

```ts
// Fulfiller holds NFT #1 and NFT #2 from the same ERC721 contract.
// Two buy orders each want one of them.
const { actions } = await seaport.fulfillOrders({
  fulfillOrderDetails: [
    { order: buyOrderForNft1 },  // consideration: ERC721 tokenId="1"
    { order: buyOrderForNft2 },  // consideration: ERC721 tokenId="2" (same contract)
  ],
  accountAddress: fulfillerAddress,
  exactApproval: true,
})

// actions only contains approve(tokenId="1").
// approve(tokenId="2") is missing — on-chain tx reverts.
```

There are no tests for `exactApproval: true` in `fulfill-orders.spec.ts`, which is why this wasn't caught.

## Fix

Replace the token-only check with `getApprovalDedupKey`, consistent with `createBulkOrders`:

```ts
// src/utils/fulfill.ts
import { getApprovalActions, getApprovalDedupKey } from "./approval"

const addApprovalIfNeeded = (
  orderInsufficientApprovals: InsufficientApprovals,
) => {
  orderInsufficientApprovals.forEach(insufficientApproval => {
    const key = getApprovalDedupKey(insufficientApproval, exactApproval)
    if (
      !totalInsufficientApprovals.some(
        approval => getApprovalDedupKey(approval, exactApproval) === key,
      )
    ) {
      totalInsufficientApprovals.push(insufficientApproval)
    }
  })
}
```

Sertug17:
Fixes #934

## Motivation

When `fulfillOrders` is called with `exactApproval: true` and multiple orders whose consideration includes ERC721 tokens from the **same contract** (different token IDs), the fulfiller's on-chain transaction silently reverts.

Root cause: `addApprovalIfNeeded` inside `fulfillAvailableOrders` deduplicates insufficient approvals by **token address only**:

```ts
!totalInsufficientApprovals.find(
  approval => approval.token === insufficientApproval.token,
)
```

For `exactApproval: false` this is harmless one `setApprovalForAll` covers all token IDs. But for `exactApproval: true`, the library must emit a separate `approve(tokenId)` per NFT. The token-only dedup discards every additional NFT from the same collection before `getApprovalActions` ever sees it.

There are no tests for `exactApproval: true` in `fulfill-orders.spec.ts`, which is why this was not caught.

## Solution

Replace the token-only comparison with `getApprovalDedupKey` the same function `createBulkOrders` already uses for approval deduplication. For `exactApproval: true` + ERC721, this returns `token:operator:identifierOrCriteria`, ensuring each token ID gets its own approval action.

3 lines changed in `src/utils/fulfill.ts`.

### DIFF
--- a/src/utils/fulfill.ts
+++ b/src/utils/fulfill.ts
@@ -23,7 +23,7 @@ import type {
   OrderWithCounter,
   SeaportContract,
 } from "../types"
-import { getApprovalActions } from "./approval"
+import { getApprovalActions, getApprovalDedupKey } from "./approval"
 import {
   type BalancesAndApprovals,
   type InsufficientApprovals,
@@ -642,9 +642,10 @@ export function fulfillAvailableOrders({
     orderInsufficientApprovals: InsufficientApprovals,
   ) => {
     orderInsufficientApprovals.forEach(insufficientApproval => {
+      const key = getApprovalDedupKey(insufficientApproval, exactApproval)
       if (
-        !totalInsufficientApprovals.find(
-          approval => approval.token === insufficientApproval.token,
+        !totalInsufficientApprovals.some(
+          approval => getApprovalDedupKey(approval, exactApproval) === key,
         )
       ) {
         totalInsufficientApprovals.push(insufficientApproval)
--- a/test/fulfill-orders.spec.ts
+++ b/test/fulfill-orders.spec.ts
@@ -1144,6 +1144,60 @@ describeWithFixture(
       })
     })
 
+
+    describe("exactApproval: true", () => {
+      it("generates a separate approve action for each ERC721 token ID when filling multiple buy orders from the same collection", async () => {
+        const { seaport, testErc721, testErc20 } = fixture
+
+        await testErc721.mint(await fulfiller.getAddress(), nftId)
+        await testErc721.mint(await fulfiller.getAddress(), nftId2)
+        await testErc20.mint(await offerer.getAddress(), parseEther("20").toString())
+        await testErc20.mint(await secondOfferer.getAddress(), parseEther("10").toString())
+
+        const firstOrder = await (
+          await seaport.createOrder({
+            offer: [{ amount: parseEther("10").toString(), token: await testErc20.getAddress() }],
+            consideration: [{ itemType: ItemType.ERC721, token: await testErc721.getAddress(), identifier: nftId, recipient: await offerer.getAddress() }],
+            fees: [{ recipient: await zone.getAddress(), basisPoints: 250 }],
+          })
+        ).executeAllActions()
+
+        const secondOrder = await (
+          await seaport.createOrder(
+            {
+              offer: [{ amount: parseEther("10").toString(), token: await testErc20.getAddress() }],
+              consideration: [{ itemType: ItemType.ERC721, token: await testErc721.getAddress(), identifier: nftId2, recipient: await secondOfferer.getAddress() }],
+              fees: [{ recipient: await zone.getAddress(), basisPoints: 250 }],
+            },
+            await secondOfferer.getAddress(),
+          )
+        ).executeAllActions()
+
+        const { actions } = await seaport.fulfillOrders({
+          fulfillOrderDetails: [{ order: firstOrder }, { order: secondOrder }],
+          accountAddress: await fulfiller.getAddress(),
+          exactApproval: true,
+        })
+
+        // With exactApproval=true, each token ID needs its own approve() call.
+        // Before the fix, addApprovalIfNeeded deduped by token only, so
+        // the approval for nftId2 was silently dropped.
+        const nftApprovals = actions.filter(
+          a => a.type === "approval" && a.token === (await testErc721.getAddress()),
+        )
+        expect(nftApprovals.length).to.equal(2)
+        expect(nftApprovals.map(a => a.identifierOrCriteria)).to.have.members([nftId, nftId2])
+
+        for (const action of actions.filter(a => a.type === "approval")) {
+          await action.transactionMethods.transact()
+        }
+
+        await actions[actions.length - 1].transactionMethods.transact()
+
+        expect(await testErc721.ownerOf(nftId)).to.equal(await offerer.getAddress())
+        expect(await testErc721.ownerOf(nftId2)).to.equal(await secondOfferer.getAddress())
+      })
+    })
     // TODO
     describe("Special use cases", () => {
       it("Can fulfill dutch auction orders", () => {})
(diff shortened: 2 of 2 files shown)


Sertug17:
Hey Sir @ryanio , this is ready for review when you get a chance.

Quick summary: addApprovalIfNeeded in fulfillAvailableOrders was deduplicating by token address only, which silently dropped approve(tokenId) calls for additional ERC721s from the same collection when exactApproval: true. The fix uses getApprovalDedupKey the same function createBulkOrders already relies on so the two code paths are now consistent. Added a regression test covering this scenario. 

If you notice anything missing, I’d be happy to correct it.

ryanio:
Thanks, the diagnosis and the source fix are both right. I confirmed against `main`: `addApprovalIfNeeded` still dedups on `approval.token === insufficientApproval.token`, and `getApprovalDedupKey` is the correct helper, already used this way in `createBulkOrders`. Swapping to it is the fix.

Two things needed before this can merge.

**The branch is 29 commits behind `main`.** `src/utils/fulfill.ts` and `src/utils/approval.ts` have both moved since you branched, so a rebase is needed.

**The test doesn't compile.** `npx tsc --noEmit` gives three errors:

```
test/fulfill-orders.spec.ts(1186,54): error TS1308: 'await' expressions are only allowed within async functions
test/fulfill-orders.spec.ts(1189,40): error TS2339: Property 'identifierOrCriteria' does not exist on type
                                       'ApprovalAction | ExchangeAction<never>'
```

The causes are `await testErc721.getAddress()` inside a non-async `filter` callback, and reading `identifierOrCriteria` off the unnarrowed action union. `filter` with a boolean predicate doesn't narrow, so you need a type predicate:

```ts
const erc721Address = await testErc721.getAddress()

const nftApprovals = actions.filter(
  (action): action is ApprovalAction =>
    action.type === "approval" && action.token === erc721Address,
)
```

with `ApprovalAction` imported from `../src/types`. Hoisting the awaited addresses above the callbacks fixes the first error and reads better anyway.

I verified the test is a genuine tripwire once it compiles: with your `fulfill.ts` change reverted it fails with `expected [ { type: 'approval' } ] to have a length of 2 but got 1`, which is exactly the dropped approval from #934. With the fix it passes, 139 passing overall.

If you'd rather not do another round, say so and I'll land the rebase and the test fix myself with you as co-author. Either way the fix is yours.

Sertug17:
Rebased onto main and fixed both TypeScript errors:
- Hoisted `await testErc721.getAddress()` above the filter callback (TS1308)
- Added `ApprovalAction` type predicate to narrow the action union (TS2339)
- Imported `ApprovalAction` from `../src/types`

CI is waiting. @ryanio

ryanio:
Thanks for turning that around fast. The rebase is right and both TypeScript errors are gone, `tsc --noEmit` is clean now. Two things still block it, and I'm sorry to send you back a third time.

**The test reverts.** It compiles but fails at runtime:

```
SolidityError: VM Exception while processing transaction: reverted with panic code 0x11
  at TestERC20.transferFrom (solmate/src/tokens/ERC20.sol:97)
  at Seaport._performERC20Transfer (seaport-core/src/lib/TokenTransferrer.sol:137)
```

I narrowed it to the `fees` entries on the two orders. What I actually tried, on your branch:

| Change | Result |
|---|---|
| your branch as-is | reverts, panic 0x11 |
| mint 20 instead of 10 to `secondOfferer` | still reverts |
| additionally mint ERC20 to `fulfiller` | still reverts |
| delete the two `fees:` lines in the new test | **passes, 139 passing** |

So it isn't a balance shortfall on any of the three participants. I haven't pinned down the exact fee-item mechanism for these offer-shaped orders, and I'd rather say so than hand you a confident wrong explanation.

The practical point: the regression you're proving is that `addApprovalIfNeeded` emits one approve action per token ID instead of collapsing them, and that doesn't depend on fees at all. Dropping them makes the test pass and say exactly what it means. That's the minimal change, deleting those two lines.

**Biome fails.** `lint` is a required check, so this fails CI on its own. The `ApprovalAction` import needs wrapping and several new lines exceed the width. `npx biome check --write test/fulfill-orders.spec.ts` handles all of it.

I ran the full check on your branch with just those two fee lines removed, and it's a genuine tripwire:

- with your `fulfill.ts` change: 139 passing
- with `fulfill.ts` reverted to `main`: fails with `AssertionError: expected 1 to equal 2`

That second line is the one that matters. It's failing on the approval count, which is the actual bug from #934, not on some incidental setup detail.

Make those two edits and it's ready to go.

Sertug17:
Fixed both issues:
- Removed the `fees` entries from both orders in the test they caused the panic 0x11 revert at `ERC20.transferFrom`. The test only needs to verify approval count, fees are not relevant.
- Ran `npx biome check --write` import wrapping and line lengths corrected.

@ryanio

ryanio:
That's it, this is ready. Verified on `96e7f27`:

- rebased onto `main`, zero commits behind
- `tsc --noEmit` clean
- `biome check` clean on both files
- full suite: 139 passing, 0 failing
- tripwire holds: with `fulfill.ts` reverted to `main` it fails with `AssertionError: expected 1 to equal 2`

Three rounds is more than I'd normally ask of a contributor, and you took each one without argument and fixed the actual thing rather than the symptom. The original diagnosis was right from the start, and the dedup-key reasoning in your first comment is what made this easy to confirm.

I've kicked off CI. Queued for merge alongside the rest of this batch, and it'll ship in the next release with you credited. Thanks also for #947, which is landing as #948.

ryanio (approved):

PR #941, opened 2026-08-11
docs: use current Alchemy RPC endpoint in README examples

doresa0:
## Summary

README examples still used deprecated `alchemyapi.io`. Switch to `g.alchemy.com`.

## Testing

- Diff-only README change


ryanio:
Taking this, thanks. `alchemyapi.io` has been the legacy host for a while and `g.alchemy.com` is correct for current keys.

Verified locally: rebased on `main`, 138 passing.

ryanio (approved):

PR #942, opened 2026-08-11
fix empty overrides handling in transaction methods

UfukNode:
## Motivation

Passing an empty overrides object currently leaves it in the contract args instead of treating it as overrides. For example, `cancelOrders(..., {})` ends up calling ethers with `[orders, {}, {}]` and fails with `no matching fragment`.

## Solution

Treat plain empty objects as valid overrides, while still avoiding false positives for primitive args and arrays. Added a regression test around `cancelOrders(..., {})`.

Validated locally with:

- `npm run build`
- `npm run lint`
- `npm run test`
- `npm run coverage`
- `npm run format:check`

### DIFF
--- a/src/utils/usecase.ts
+++ b/src/utils/usecase.ts
@@ -61,11 +61,15 @@ const instanceOfOverrides = <T extends Overrides>(
     "overrides",
   ]
 
-  return (
-    obj === undefined ||
-    (Object.keys(obj).length > 0 &&
-      Object.keys(obj).every(key => validKeys.includes(key)))
-  )
+  if (obj === undefined) {
+    return true
+  }
+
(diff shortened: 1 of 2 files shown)


ryanio:
Taking this. Confirmed the cause: `Object.keys(obj).length > 0` meant `{}` failed the overrides check and got passed through as a contract argument, producing the `no matching fragment` error you described.

I also checked the replacement doesn't widen anything by accident. Primitives and arrays returned `false` before (an empty array hit the length check, a string's index keys failed `validKeys`) and still return `false` now via the explicit guards, so `{}` is the only input whose result changes.

Verified locally: rebased on `main`, types clean, 139 passing.

ryanio (approved):

PR #943, opened 2026-08-11
fix(utils): normalize negative inputs in gcd and add empty array guard to findGcd

yappermoar-boop:
Updated gcd and findGcd in src/utils/gcd.ts. Added an empty array check in findGcd to prevent throwing TypeError: Cannot convert undefined to a BigInt when passed an empty array. Also normalized inputs in gcd to absolute values to guarantee non-negative GCD outputs per mathematical definition. Added unit test coverage.

### DIFF
--- a/src/utils/gcd.ts
+++ b/src/utils/gcd.ts
@@ -1,8 +1,11 @@
 import type { BigNumberish } from "ethers"
 
 export const gcd = (a: BigNumberish, b: BigNumberish): bigint => {
-  const bnA = BigInt(a)
-  const bnB = BigInt(b)
+  let bnA = BigInt(a)
+  let bnB = BigInt(b)
+
+  bnA = bnA < 0n ? -bnA : bnA
+  bnB = bnB < 0n ? -bnB : bnB
 
   if (bnA === 0n) {
     return bnB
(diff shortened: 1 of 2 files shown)


ryanio:
Thanks for looking at this. I traced both changes against the callers and I don't think either one is reachable, so I'm going to pass.

`findGcd` has exactly two callers: `getMaximumSizeForOrder`, which passes `startAmount`/`endAmount` from an order's offer and consideration items, and `fulfill.ts` line 911. Those amounts are unsigned integer strings from order parameters, so they're never negative, and a valid order always has at least one offer item, so the array is never empty.

On the empty case specifically, returning `0n` picks a semantic rather than avoiding one. `getMaximumSizeForOrder` returning `0` units would flow into the divisibility check and produce a confusing downstream error instead of the `TypeError` you're avoiding. If we ever do need to handle an empty array here, throwing with a clear message is probably the more honest choice.

Not a criticism of the change itself, the absolute-value normalization is correct as mathematics. It's just guarding inputs the type and the call sites don't produce, and I'd rather not add branches that no test can exercise for real. Appreciate you taking the time, and please do keep them coming, the other reports in this batch found real bugs.

PR #944, opened 2026-08-11
fix: sum token amounts case-insensitively in getSummedTokenAndIdentifierAmounts

memosr:
## Problem

`getSummedTokenAndIdentifierAmounts` keys its accumulator on the raw `item.token` string:

```ts
[item.token]: {
  ...map[item.token],
```

Token addresses reach this function from several sources. Order items typically arrive lowercased from an API, while `tips` are constructed by the caller and are commonly checksummed. When the same token appears in both casings, the amounts land in two separate buckets.

`getInsufficientBalanceAndApprovalAmounts` then walks those buckets and compares each one against the owner's full balance independently. An order requiring 100 + 200 of the same ERC20 from a wallet holding 250 passes the check, because neither bucket alone exceeds the balance. The revert surfaces on-chain instead of in the pre-flight check the library exists to provide.

## Fix

Normalize the accumulator key with `toLowerCase()`. This matches the behaviour `findBalanceAndApproval` already relies on, which compares both `token` and `identifierOrCriteria` case-insensitively, so the summation and the lookup now agree.

Native-token lookups are unaffected: callers read the map at `ethers.ZeroAddress`, which contains no alphabetic characters.

## Test

Added a unit test in `test/utils/item.spec.ts` covering the same ERC20 supplied in checksummed and lowercased form. It fails on `main` (two keys, `100n` and `200n`) and passes with the fix (one key, `300n`).

Full suite: 139 passing.

### DIFF
--- a/src/utils/item.ts
+++ b/src/utils/item.ts
@@ -106,15 +106,19 @@ export const getSummedTokenAndIdentifierAmounts = ({
   >((map, item) => {
     const identifierOrCriteria =
       itemToCriteria.get(item)?.identifier ?? item.identifierOrCriteria
+    // The same token can arrive in different casings (e.g. checksummed from one
+    // source, lowercased from another). Normalize the key so those items are
+    // summed into a single bucket, matching the case-insensitive lookup done by
+    // findBalanceAndApproval.
+    const token = item.token.toLowerCase()
 
     return {
       ...map,
-      [item.token]: {
-        ...map[item.token],
+      [token]: {
+        ...map[token],
         // Being explicit about the undefined type as it's possible for it to be undefined at first iteration
         [identifierOrCriteria]:
-          ((map[item.token]?.[identifierOrCriteria] as bigint | undefined) ??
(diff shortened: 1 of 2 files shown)


ryanio:
Taking this. I checked the two things that could have made lowercasing the key unsafe, and both are fine:

- `findBalanceAndApproval` already compares `token.toLowerCase()` on both sides, so the lookup agrees with the normalized key, as you said.
- The other readers of the map index it at `ethers.ZeroAddress` for the native total, which has no alphabetic characters, so they're unaffected.

The split-bucket consequence is the part that makes this worth fixing: `getInsufficientBalanceAndApprovalAmounts` checks each bucket against the full balance independently, so two casings of the same ERC20 both pass while the sum doesn't. That turns a pre-flight check into an onchain revert, which is the opposite of what this library is for.

Verified locally: rebased on `main`, types clean, 139 passing.

ryanio (approved):

PR #945, opened 2026-08-12
fix: encode buffers as hex without BigNumberish coercion

crazywriter1:
## Summary

- Fix `bufferToHex` to prefix buffer hex with `0x` instead of passing a bare hex string to `toBeHex`
- Add unit tests for `0x` prefixing, full-width zero buffers, and round-trip with `hexToBuffer`
- 
## Problem

`bufferToHex` used `toBeHex(buf.toString("hex"))`. ethers treats a hex string without `0x` as a BigNumberish decimal string, so non-zero buffers throw:
    invalid BigNumberish string: Cannot convert abcd to a BigInt
This breaks `getBulkOrderHash()`, which depends on `bufferToHex` for the merkle root.

## Test plan

- [x] npm run lint
- [x] npm test -- --grep "eip712 utils"

### DIFF
--- a/src/utils/eip712/utils.ts
+++ b/src/utils/eip712/utils.ts
@@ -1,5 +1,5 @@
 import type { BytesLike } from "ethers"
-import { concat, keccak256, toBeHex } from "ethers"
+import { concat, keccak256 } from "ethers"
 
 export const makeArray = <T>(len: number, getValue: (i: number) => T) =>
   Array(len)
@@ -12,7 +12,7 @@ export const chunk = <T>(array: T[], size: number) => {
   )
 }
 
-export const bufferToHex = (buf: Buffer) => toBeHex(buf.toString("hex"))
+export const bufferToHex = (buf: Buffer) => `0x${buf.toString("hex")}`
 
 export const hexToBuffer = (value: string) => Buffer.from(value.slice(2), "hex")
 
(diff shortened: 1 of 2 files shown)


ryanio:
The fix is correct and I'm taking it, but I want to correct the impact claim so it doesn't end up in the changelog that way.

You're right about the mechanism. I confirmed it directly:

```
toBeHex("abcd")  ->  THROWS: invalid BigNumberish string: Cannot convert abcd to a BigInt
```

And there's a second failure mode worth naming: a 64-character hash that happens to be all decimal digits doesn't throw, it parses as a decimal number and returns a silently wrong, shorter hex string. That's worse than the throw.

Where I'd push back is "This breaks `getBulkOrderHash()`". `getBulkOrderHash` has no callers anywhere in `src` or `test`. `bufferToHex` is reachable only from it, at `Eip712MerkleTree.ts` line 101. So bulk order creation is not broken today, and no live signature or merkle root changes as a result of this fix. That's also why it's zero-risk to take.

Net: real bug, correct fix, currently unreachable. Taking it as a correctness fix with the tests, and it'll ship described that way rather than as a bulk-order fix.

Verified locally: rebased on `main`, types clean, 141 passing.

crazywriter1:
Thanks for the review sir. Sounds good.

ryanio (approved):

PR #946, opened 2026-08-12
fix: account for tips when choosing the basic fulfill route

memosr:
## Problem

`fulfillOrder` picks between the basic and standard routes with:

```ts
shouldUseBasicFulfill(sanitizedOrder.parameters, totalFilled)
```

`tipConsiderationItems` is built a few lines above this call, but is not passed in. Everywhere else in the same flow tips are treated as consideration: `fulfillBasicOrder` appends them to `additionalRecipients`, computes `totalNativeAmount` over `considerationIncludingTips`, and `validateBasicFulfillBalancesAndApprovals` validates against `[...consideration, ...tips]`. The route decision is the one place they are dropped.

Seaport pays every entry in `additionalRecipients` in `considerationToken`. So for a native listing fulfilled with an ERC20 tip, `shouldUseBasicFulfill` still returns `true`, and the resulting `fulfillBasicOrder` call asks Seaport to pay the tip recipient in ETH rather than the ERC20 the caller specified. Rule 5 (all currencies must match) is exactly the guard that should have caught this, but it only ever saw the order's own consideration.

The same gap applies to the other consideration-wide rules: a tip with `startAmount !== endAmount` (rule 6), or one that pushes the additional-recipient total past the offered amount on a same-currency order (rule 8), slips through as well.

## Fix

`shouldUseBasicFulfill` now takes the tips as an optional third argument and applies rules 5, 6, 8 and 9 to `[...consideration, ...tips]`. Rule 2 still checks the order's own consideration, since a tip cannot stand in for a missing consideration item. When the combined set does not satisfy the basic-order constraints, fulfillment falls back to the standard route, which handles mixed currencies correctly.

The parameter defaults to `[]`, so existing callers are unaffected.

## Test

Added three cases to `test/fulfill-utils.spec.ts`:

- a plain native listing still takes the basic route
- a tip in the order's own currency still takes the basic route
- an ERC20 tip on a native listing does not

The third fails on `main` (`expected true to equal false`) and passes with the fix.

Full suite: 141 passing. `tsc --noEmit` clean.

### DIFF
--- a/src/seaport.ts
+++ b/src/seaport.ts
@@ -884,7 +884,11 @@ export class Seaport {
     if (
       !unitsToFill &&
       isRecipientSelf &&
-      shouldUseBasicFulfill(sanitizedOrder.parameters, totalFilled)
+      shouldUseBasicFulfill(
+        sanitizedOrder.parameters,
+        totalFilled,
+        tipConsiderationItems,
+      )
     ) {
       // TODO: Use fulfiller proxy if there are approvals needed directly, but none needed for proxy
       return fulfillBasicOrder(
--- a/src/utils/fulfill.ts
+++ b/src/utils/fulfill.ts
@@ -74,6 +74,7 @@ import {
 export const shouldUseBasicFulfill = (
   { offer, consideration, offerer }: OrderParameters,
   totalFilled: OrderStatus["totalFilled"],
+  tips: ConsiderationItem[] = [],
 ) => {
   // 1. The order must not be partially filled
   if (totalFilled !== 0n) {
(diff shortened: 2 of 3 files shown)


ryanio:
This is a good catch and the reasoning holds up. I verified the part that would have made it unsafe: rule 2 guarantees `consideration.length >= 1`, so pulling `firstConsideration` from the combined array can't change which item rule 7 checks. And you're right that the route decision is the one place in that flow where tips are dropped, since `fulfillBasicOrder`, `totalNativeAmount`, and `validateBasicFulfillBalancesAndApprovals` all already treat them as consideration.

Verified locally: rebased on `main`, types clean, 141 passing, and your third case does fail on `main` as described.

I want one more pass on this before it lands, because it changes route selection rather than just rejecting bad input: orders that took the basic route before will now take the standard route in more cases. That's the correct outcome for the mixed-currency case, but I'd like to confirm nothing downstream of us depends on basic-route selection with tips present. Not a concern about correctness, just the one change in this batch with a behavioural blast radius beyond the bug it fixes.

Marking it for a minor release alongside #940.

ryanio (approved):

PR #948, opened 2026-08-12
fix: use nullish check for overrides.value in _validateMatchOrdersNativeValue

### LINKED ISSUE #947: matchOrders / matchAdvancedOrders incorrectly throws when overrides.value is explicitly set to 0n
### Component

API

- [x] seaport-js
- [x] Node (minimum v16)

4.1.7

_validateMatchOrdersNativeValue

### Operating System

Linux

## Summary

`matchOrders` and `matchAdvancedOrders` throw the error
_"Orders contain native ETH items but no `value` was provided in overrides"_
even when `overrides.value` is **explicitly set to `0n`**.

The root cause is a falsy check on `overrides?.value`. In JavaScript/TypeScript,
`BigInt(0)` (`0n`) is falsy, so `!overrides?.value` evaluates to `true` when
`value` is `0n` the same result as when `value` is `undefined`.

## Root cause

```ts
// src/seaport.ts _validateMatchOrdersNativeValue
if (hasNativeToken && !overrides?.value) {   // ← !0n === true  (bug)
  throw new Error(
    "Orders contain native ETH items but no `value` was provided in overrides. ..."
  )
}
```

| `overrides?.value` | `!overrides?.value` | Should throw? |
|---|---|---|
| `undefined` | `true` | ✅ yes — value was forgotten |
| `null` | `true` | ✅ yes |
| `0n` | **`true`** | ❌ **no — value was explicitly provided** |
| `1000000n` | `false` | ✅ no |

## Reproduction

```ts
// Both matchOrders and matchAdvancedOrders are affected.
// Passing value: 0n (explicitly saying "no ETH needed from me") still throws.
seaport.matchOrders({
  orders,
  fulfillments,
  overrides: { value: 0n },   // ← explicit zero but still throws!
})
// Error: "Orders contain native ETH items but no `value` was provided in overrides."
```

This is misleading: the error message tells the user to pass `value` but they
already did. The check cannot distinguish between a forgotten value and an
intentional zero.

## Fix

Replace the falsy check with a nullish check:

```ts
// Before:
if (hasNativeToken && !overrides?.value) {

// After:
if (hasNativeToken && overrides?.value == null) {
```

`== null` is `true` only for `undefined` and `null` not for `0n` or `0`.
One-line change. I'll open a PR with the fix and a regression test.

Sertug17:
## Summary

Fixes #947

`_validateMatchOrdersNativeValue` used `!overrides?.value` to detect a missing ETH value. Since `BigInt(0)` (`0n`) is falsy in JavaScript, this check incorrectly throws when `overrides: { value: 0n }` is explicitly passed.

## Root cause

```ts
// Before:
if (hasNativeToken && !overrides?.value) {   // !0n === true  ← bug

// After:
if (hasNativeToken && overrides?.value == null) {  // true only for undefined/null
```

## Changes

- `src/seaport.ts`: one-line fix — `!overrides?.value` → `overrides?.value == null`
- `test/match-orders.spec.ts`: regression test verifying `value: 0n` does not throw

### DIFF
--- a/src/seaport.ts
+++ b/src/seaport.ts
@@ -1260,7 +1260,7 @@ export class Seaport {
         params.consideration.some(item => item.itemType === ItemType.NATIVE),
     )
 
-    if (hasNativeToken && !overrides?.value) {
+    if (hasNativeToken && overrides?.value == null) {
       throw new Error(
         "Orders contain native ETH items but no `value` was provided in overrides. " +
           "Pass the required ETH amount via `overrides: { value }` to avoid a revert.",
--- a/test/match-orders.spec.ts
+++ b/test/match-orders.spec.ts
@@ -133,6 +133,34 @@ describeWithFixture("As a user I want to match an order", fixture => {
             "Orders contain native ETH items but no `value` was provided",
           )
         })
+
+        it("does not throw when value is explicitly 0n for ETH orders", async () => {
+          const { seaport } = fixture
+
+          const { executeAllActions } = await seaport.createOrder(
+            privateListingCreateOrderInput,
+          )
+
+          const order = await executeAllActions()
+
+          const recipientAddress = await privateListingRecipient.getAddress()
+          const counterOrder = constructPrivateListingCounterOrder(
+            order,
+            recipientAddress,
+          )
+          const fulfillments = getPrivateListingFulfillments(order)
+
+          // Regression: !overrides?.value treats 0n as falsy and incorrectly
+          // throws. overrides?.value == null must be used instead.
+          expect(() =>
+            seaport.matchOrders({
+              orders: [order, counterOrder],
+              fulfillments,
+              overrides: { value: 0n },
+              accountAddress: recipientAddress,
+            }),
+          ).not.to.throw()
+        })
       })
 
       describe("with ERC20", () => {
--- a/test/match-orders.spec.ts
+++ b/test/match-orders.spec.ts
@@ -161,6 +161,42 @@ describeWithFixture("As a user I want to match an order", fixture => {
             }),
           ).not.to.throw()
         })
+
+        it("does not throw when value is explicitly 0n for ETH orders (matchAdvancedOrders)", async () => {
+          const { seaport } = fixture
+
+          const { executeAllActions } = await seaport.createOrder(
+            privateListingCreateOrderInput,
+          )
+
+          const order = await executeAllActions()
+
+          const recipientAddress = await privateListingRecipient.getAddress()
+          const counterOrder = constructPrivateListingCounterOrder(
+            order,
+            recipientAddress,
+          )
+          const fulfillments = getPrivateListingFulfillments(order)
+
+          // Regression: same falsy-0n bug exists in matchAdvancedOrders via
+          // the shared _validateMatchOrdersNativeValue helper.
+          expect(() =>
+            seaport.matchAdvancedOrders({
+              orders: [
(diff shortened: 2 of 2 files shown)


ryanio:
Verified and taking this. Fast turnaround, thanks.

Checked the one thing that would have made the test vacuous: `matchOrders` is declared `public matchOrders({` with no `async`, so `_validateMatchOrdersNativeValue` throws synchronously and `expect(...).not.to.throw()` is a valid assertion here. If the method were async the throw would surface as a rejected promise and the test would pass either way.

Confirmed failing-first: with `overrides?.value == null` reverted to `!overrides?.value`, the test fails. With the fix, 139 passing, types and Biome clean, rebased on `main`.

One follow-up rather than a change request: `matchAdvancedOrders` shares the same helper, so it's covered by the same one-line fix, but the test only exercises `matchOrders`. Not going to hold the fix for it. Worth adding if you're back in this file.

Sertug17:
matchAdvancedOrders regression already added in the second commit (8ec69a0) both methods covered. @ryanio

ryanio:
Thanks for adding the `matchAdvancedOrders` case, that was the right call. One thing to fix: CI's `lint` is failing, and it's `check-types` inside it rather than Biome.

This is a case where a green test run is misleading. Mocha runs through hardhat without typechecking, so the new test passes at runtime while `tsc --noEmit` rejects it. `npm test` alone won't show you this; `npm run lint` will.

Four errors, two separate causes.

`numerator` and `denominator` are `bigint`, not `number`:

```
test/match-orders.spec.ts(186,29): error TS2322: Type 'number' is not assignable to type 'bigint'
test/match-orders.spec.ts(189,19): error TS2322: Type 'number' is not assignable to type 'bigint'
```

So `1n` in both objects, including the spread one on line 186.

Then, once those are fixed, a fourth error appears that the first two were masking: `matchAdvancedOrders` requires `criteriaResolvers` and `recipient`, which `matchOrders` does not.

```
error TS2345: ... is missing the following properties: criteriaResolvers, recipient
```

The full set of edits to the `matchAdvancedOrders` call:

```ts
orders: [
  { ...order, numerator: 1n, denominator: 1n, extraData: "0x" },
  { ...counterOrder, numerator: 1n, denominator: 1n, extraData: "0x" },
],
criteriaResolvers: [],
fulfillments,
recipient: recipientAddress,
overrides: { value: 0n },
accountAddress: recipientAddress,
```

Verified locally with exactly that: `tsc --noEmit` clean, `biome check .` clean across all 53 files, 140 passing. And both cases are real tripwires, with the `== null` change reverted, both fail rather than just the `matchOrders` one.

Worth running `npm run lint` and not only `npm test` before the next push, since that's the check that would have caught this.

Sertug17:
Fixed both TypeScript errors:
- `numerator` and `denominator` changed to `1n` (bigint)
- Added `criteriaResolvers: []` and `recipient` to the `matchAdvancedOrders` call

@ryanio

ryanio:
Verified on `2e87cf3`, this is ready:

- `npm run lint` clean (both `check-types` and `biome check .` across 53 files)
- 140 passing, 0 failing
- rebased, zero commits behind `main`
- both cases are real tripwires: with the `== null` change reverted, `matchOrders` **and** `matchAdvancedOrders` both fail

CI is green on all three required checks.

Nice work on this one, you found it, reported it with the right diagnosis and a correct fix, and turned around two review rounds inside an hour. Queued for merge with the rest of the batch.

ryanio (approved):

PR #951, opened 2026-08-12
fix: do not mutate caller's CreateOrderInput in createBulkOrders

### LINKED ISSUE #949: createBulkOrders mutates caller's CreateOrderInput objects via ??= assignment
### Component

API

- [x] seaport-js
- [x] Node (minimum v16)

4.1.7

createBulkOrders

### Operating System

Linux

## Summary

`createBulkOrders` mutates the caller's `CreateOrderInput` objects in place. On every call it writes back to `input.counter` via the `??=` operator,
permanently modifying an object the caller owns and did not expect to change.

## Root cause

```ts
// src/seaport.ts — createBulkOrders, inside the for-loop
for (const input of createOrderInput) {
  input.counter ??= offererCounter   // ← mutates the caller's object
```

`??=` assigns `offererCounter` to `input.counter` when it is `null` or
`undefined`. Because `input` is a direct reference into the caller-supplied
array, the original object is modified.

## Reproduction

```ts
const myOrder: CreateOrderInput = {
  offer: [{ itemType: ItemType.ERC721, token: nftAddress, identifier: "1" }],
  consideration: [{ amount: parseEther("1"), recipient: offerer }],
}

console.log(myOrder.counter) // undefined caller never set it

await seaport.createBulkOrders([myOrder], offererAddress)

console.log(myOrder.counter) // now holds the on-chain counter value!
                              // caller's object was silently modified
```

This is surprising: the caller passed an input configuration object and did not expect it to be written to. Callers who reuse or inspect the same
`CreateOrderInput` object after the call (e.g. retrying, logging, or comparing state) will see stale or unexpected data.

## Fix

Spread the input into a local copy before writing `counter`, leaving the caller's object untouched:

```ts
// Before:
input.counter ??= offererCounter
const { orderComponents, approvalActions } = await this._formatOrder(
  signer, offerer, Boolean(exactApproval), input,
)

// After:
const orderInput = { ...input, counter: input.counter ?? offererCounter }
const { orderComponents, approvalActions } = await this._formatOrder(
  signer, offerer, Boolean(exactApproval), orderInput,
)
```

Two line change. I'll open a PR with the fix and a regression test.

Sertug17:
## Summary

Fixes #949

`createBulkOrders` used `input.counter ??= offererCounter` inside a for-loop, which writes the resolved counter value back to the caller's original object.
Callers who inspect or reuse the same `CreateOrderInput` after the call see a `counter` value they never set.

## Root cause

```ts
// Before:
for (const input of createOrderInput) {
  input.counter ??= offererCounter  // ← mutates caller's object
```

`input` is a direct reference into the caller-supplied array. `??=` assigns when the left side is `null` or `undefined`, so the caller's object is permanently modified on every call.

## Fix

```ts
// After:
const orderInput = { ...input, counter: input.counter ?? offererCounter }
```

Spread into a local copy the original object is never touched.

## Changes

- `src/seaport.ts`: 2-line change (remove `??=`, add spread)
- `test/create-bulk-orders.spec.ts`: regression test asserting `counter` 
  remains `undefined` on the caller's object after `createBulkOrders` returns

### DIFF
--- a/src/seaport.ts
+++ b/src/seaport.ts
@@ -245,12 +245,12 @@ export class Seaport {
     const allOrderComponents: OrderComponents[] = []
 
     for (const input of createOrderInput) {
-      input.counter ??= offererCounter
+      const orderInput = { ...input, counter: input.counter ?? offererCounter }
       const { orderComponents, approvalActions } = await this._formatOrder(
         signer,
         offerer,
         Boolean(exactApproval),
-        input,
+        orderInput,
       )
 
       allOrderComponents.push(orderComponents)
--- a/test/create-bulk-orders.spec.ts
+++ b/test/create-bulk-orders.spec.ts
@@ -188,5 +188,39 @@ describeWithFixture(
         conduitOperator,
       ])
     })
+
+  it("does not mutate caller's CreateOrderInput objects", async () => {
+    const { seaport, testErc721, ethers } = fixture
+
+    const [offerer, zone] = await ethers.getSigners()
+
+    await testErc721.mint(await offerer.getAddress(), "1")
+
+    const input = {
+      offer: [
+        {
+          itemType: ItemType.ERC721,
+          token: await testErc721.getAddress(),
+          identifier: "1",
+        },
+      ],
+      consideration: [
+        {
+          amount: parseEther("1").toString(),
+          recipient: await offerer.getAddress(),
+        },
+      ],
+      fees: [{ recipient: await zone.getAddress(), basisPoints: 250 }],
+    }
+
+    // counter is intentionally absent — caller did not set it
+    expect((input as Record<string, unknown>).counter).to.be.undefined
+
+    await seaport.createBulkOrders([input], await offerer.getAddress())
+
+    // Regression: input.counter ??= offererCounter mutated the caller's object.
+    // After the fix (spread copy), the original must remain untouched.
+    expect((input as Record<string, unknown>).counter).to.be.undefined
+  })
   },
 )


ryanio:
Good find, and the diagnosis is right. `input.counter ??= offererCounter` writes into an object the caller owns, and the spread copy is the correct fix. Confirmed on current `main`.

Two things to fix, both in the test.

**`check-types` fails.** The object literal isn't assignable to `CreateOrderInput`, because TS widens `itemType` and the item shapes rather than inferring the union member:

```
test/create-bulk-orders.spec.ts(219,37): error TS2322: Type '{ offer: ...; consideration: ...; fees: ... }'
  is not assignable to type 'CreateOrderInput'
```

Annotate the variable and the inference lands correctly:

```ts
const input: CreateOrderInput = {
```

with `CreateOrderInput` added to the existing `import type { ... } from "../src/types"`. That also lets you drop both `as Record<string, unknown>` casts, since `counter` is a real optional property on the type, so `expect(input.counter).to.be.undefined` type-checks directly and reads better.

**Biome format fails.** The new `it` block is indented two spaces short of the surrounding block, so it lands outside the `describe` body's indentation. `npx biome check --write test/create-bulk-orders.spec.ts` fixes it.

**Also needs a rebase**, it's 9 commits behind and `src/seaport.ts` moved underneath it (#946, #948, #950 all touched that file). It rebases cleanly, no conflicts.

I verified all of the above locally: with the annotation, the import, and Biome run, `npm run lint` is clean and your test passes. Worth running `npm run lint` rather than just `npm test` before pushing, since `check-types` lives in there and the test suite won't catch it.

One unrelated note so it doesn't confuse you: on a rebase you'll currently see one failure in `fulfill-orders.spec.ts` about approval counts. That's a regression on `main` from #944 and #935 interacting, not anything you did. Fix is up in #952.

ryanio:
Landed via #953, which is merged. Your commit went in as the first of the two there, so the fix and the credit are yours, plus a co-author trailer on the squash.

I couldn't push to your fork branch from my side, so rather than send you a fourth round for what amounted to a type annotation, I carried it over: added `CreateOrderInput` to the type import and annotated the literal (which let both `as Record<string, unknown>` casts go, since `counter` is a real optional property), ran Biome on the block, and rebased onto current main.

Closing this in favour of #953. Thanks for the report in #949 and the fix, and for the three other issues you found tonight. Four real bugs in one evening is a good haul.

PR #957, opened 2026-08-13
test: add fast, isolated unit coverage for freezeOrderComponents

batuhankocyigit:
### Motivation

`freezeOrderComponents` (added recently to guarantee the `orderComponents`
exposed on a create action can't drift from what actually gets signed — see
its JSDoc in `src/utils/order.ts`) currently only has indirect coverage
through `test/execute-approvals.spec.ts`, which needs the full Hardhat
network (contract deployment, a signer, etc.) to run.

That's appropriate for testing the end-to-end feature, but the freezing
logic itself is a pure function with no on-chain dependency, and it's a
mutation-safety guarantee the JSDoc goes out of its way to explain the
reasoning for — exactly the kind of thing worth pinning down with a fast,
isolated unit test that runs in milliseconds and fails with a precise
message if the freezing behavior ever regresses.

### What this adds

A `freezeOrderComponents` describe block in `test/utils/order.spec.ts`
(alongside the existing `deductFees` suite, since both live in
`src/utils/order.ts`), covering:

- Returns the same object reference rather than a clone (the JSDoc is
  explicit that this is intentional, not an oversight).
- Freezes the top-level `OrderComponents` object.
- Deep-freezes `offer` items individually, not just the array wrapper.
- Deep-freezes `consideration` items individually, and that the array
  itself rejects mutation (e.g. `push`).
- Doesn't throw when a nested item was already frozen before the call
  (the `Object.isFrozen` short-circuit in `deepFreeze`).

### Testing

- `npx tsc --noEmit` — clean.
- `npx biome check test/utils/order.spec.ts` — clean.
- Ran the new suite directly: 9/9 passing (4 pre-existing `deductFees`
  cases + 5 new `freezeOrderComponents` cases), all in single-digit
  milliseconds since none of it touches the network.

### Scope

Test-only change, one file. No production code touched.

### DIFF
--- a/test/utils/order.spec.ts
+++ b/test/utils/order.spec.ts
@@ -1,7 +1,7 @@
 import { expect } from "chai"
-import { ItemType } from "../../src/constants"
-import type { ConsiderationItem } from "../../src/types"
-import { deductFees } from "../../src/utils/order"
+import { ItemType, OrderType } from "../../src/constants"
+import type { ConsiderationItem, OrderComponents } from "../../src/types"
+import { deductFees, freezeOrderComponents } from "../../src/utils/order"
 
 // deductFees subtracts the summed fee basisPoints from every currency item and
 // is the point where an out-of-range fee first turns an order malformed: a total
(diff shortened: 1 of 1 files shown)


ryanio (approved):
thanks!

PR #958, opened 2026-08-14
fix: handle empty domain registry in getDomain

Radovenchyk:
Fixed a bug in DomainRegistry.getDomain where querying an unregistered tag caused a Solidity Panic(0x11) due to an array length underflow instead of returning the documented DomainIndexOutOfRange error. The fix performs a safe bounds check without changing the error ABI and adds regression tests for unknown tags and out-of-range indices. All 166 tests pass and the fix was committed and pushed in d7fdcb4

### DIFF
--- a/src/contracts/DomainRegistry.sol
+++ b/src/contracts/DomainRegistry.sol
@@ -121,13 +121,16 @@ contract DomainRegistry is DomainRegistryInterface {
         view
         returns (string memory domain)
     {
-        // Get the maximum possible index of the array of registered domains for the input tag.
-        uint256 maxIndex = _registry[tag].length - 1;
+        // Get the length of the array of registered domains for the input tag.
+        uint256 length = _registry[tag].length;
 
-        // Revert if the index parameter is out of range for the array of domains
-        // corresponding to the tag.
-        if (index > maxIndex) {
-            revert DomainIndexOutOfRange(tag, maxIndex, index);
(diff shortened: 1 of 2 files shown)


ryanio:
Thanks for this, and for the clear write-up. The diagnosis is right: `getDomain` computes `_registry[tag].length - 1` before it bounds-checks, so an unregistered tag underflows and reverts with `Panic(0x11)` instead of `DomainIndexOutOfRange`. I confirmed it against the deployed registry on mainnet.

The catch is where the fix lands. `DomainRegistry` is already deployed at `0x000000000DaD0DE04D2B2D4a5A74581EBA94124A` and immutable, and we aren't redeploying it. The `.sol` in this repo is only compiled and deployed by the test fixture (`test/utils/setup.ts:102`); in production the SDK connects to that fixed address (`src/constants.ts:129`). So changing the contract source would make the repo diverge from the deployed bytecode, and the new "unregistered tag" test would pass locally while every real caller still hit the panic.

#959 does it SDK-side instead and has merged, with you credited as co-author since the diagnosis is yours. `getDomain` now catches the revert and, only when the tag has no domains at all, throws an error saying so. A tag that does have domains already reverts with `DomainIndexOutOfRange` (verified onchain), so that path is untouched.

Closing this one in favor of #959. Appreciate you taking the time to dig into it.

Radovenchyk:
Thank you for the help

PR #960, opened 2026-08-14
Improve missing criteria error context

kejan2514:
## Motivation

When criteria-based items are missing their corresponding criteria, the current error message does not provide enough context to identify which order or item caused the failure.

This can make debugging invalid or incomplete orders unnecessarily difficult.

## Solution

Improve the missing-criteria error by including contextual information about the affected order, item, and side.

This keeps the existing validation behavior while making failures easier to diagnose.

## Testing

Added tests covering missing criteria errors and verifying that the relevant context is included in the error message.

### DIFF
--- a/src/seaport.ts
+++ b/src/seaport.ts
@@ -1231,8 +1231,28 @@ export class Seaport {
    * @param index The index.
    * @returns The domain at the index for the given tag.
    */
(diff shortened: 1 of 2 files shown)


ryanio (changes requested):
thanks! just needs a lint format

PR #961, opened 2026-08-16
fix: validate individual fee basis points

wodesiku:
## Summary
- reject every fee `basisPoints` value that is negative, fractional, unsafe, or above 10,000
- keep the existing aggregate 100% cap for otherwise valid fee arrays
- add regression coverage for negative and fractional inputs

## Problem
`deductFees` previously only checked the sum of `basisPoints`. A negative fee could offset an oversized fee, allowing invalid fee values through order construction. Fractional values were also accepted even though Seaport expects integer basis points. Both paths can produce malformed consideration amounts and defer failure to ABI encoding or contract execution.

## Test plan
- `npm test` (174 passing)
- `npm run check-types`
- `npm run build`
- `npx biome check src/utils/order.ts test/utils/order.spec.ts`

## Note
`npm run lint` reaches the same TypeScript check successfully, but the repository-wide Biome phase reports pre-existing CRLF formatting diagnostics across unmodified files in this Windows checkout.

### DIFF
--- a/src/utils/order.ts
+++ b/src/utils/order.ts
@@ -85,6 +85,18 @@ export const deductFees = <T extends Item>(
     return items
   }
 
+  for (const { basisPoints } of fees) {
+    if (
+      !Number.isSafeInteger(basisPoints) ||
+      basisPoints < 0 ||
+      basisPoints > Number(ONE_HUNDRED_PERCENT_BP)
+    ) {
+      throw new Error(
+        `Fee basisPoints (${basisPoints}) must be a safe integer between 0 and ${ONE_HUNDRED_PERCENT_BP} (100%).`,
+      )
+    }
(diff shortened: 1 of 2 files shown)


ryanio:
Taken, thanks. A negative or fractional `basisPoints` went straight into `multiplyBasisPoints` before this, since only the total was bounded.

One note so it lands correctly in the changelog. Because your per-fee check runs before the total check, a single fee above 10000 now reports "Fee basisPoints (10001) must be a safe integer..." rather than the old "Total fee basisPoints..." message, which is why the existing test needed updating. A new throw on input that previously returned makes this a minor release rather than a patch.

Landing in #972 with the other contributions that touched these files. Your commit is in there authored by you. Closing in favor of that.

PR #962, opened 2026-08-16
fix: include tips in the fulfiller's balance and approval checks

Dusk1e:
## Problem

`fulfillOrder` and `fulfillOrders` build the fulfiller's balance and approval set from the order's own items only:

```ts
items: [...offer, ...consideration]
```

The checks that read that set sum tips in alongside the consideration. So a tip in a token the order does not already carry has no entry to look up, and `findBalanceAndApproval` throws before any transaction is built:

```
Checking for balance and approvals for token 0x0000000000000000000000000000000000000000 id 0 failed
```

That rules out a native tip on an ERC20 listing, a tip in a second ERC20, and NFT tips, all of which Seaport itself accepts. Tipping in the same token as the order works fine, which is why the existing tip coverage in `partial-fulfill.spec.ts` never hit this.

## Fix

Include the mapped tip items in the fulfiller's lookup set in both entry points. They go on the end, after the order's items, so the offer and consideration criteria mapping in `getItemToCriteriaMap` is untouched.

It also makes the check do its job rather than just stop crashing: a tip in a token the fulfiller has not approved now produces its own approval action, which the fulfillment would revert without.

In `fulfillOrders` the per-order tip mapping is hoisted so it is built once and reused for both the lookup set and `ordersMetadata`, instead of mapped twice.

## Test

`test/tips.spec.ts` covers both entry points. Both fail on `main` with the error above.

- `fulfillOrder` with a second-ERC20 tip: asserts both approval actions appear, then fulfills and checks the NFT and the tip both land.
- `fulfillOrders` with a native tip on an ERC20 listing: asserts the tip rides as `msg.value`, then fulfills and checks the recipient's balance moved.

Suite green at 174.

### DIFF
--- a/src/seaport.ts
+++ b/src/seaport.ts
@@ -839,6 +839,11 @@ export class Seaport {
 
     const fulfillerOperator = this.config.conduitKeyToConduit[conduitKey]
 
+    const tipConsiderationItems = tips.map(tip => ({
+      ...mapInputItemToOfferItem(tip),
+      recipient: tip.recipient,
+    }))
+
     const [
       offererBalancesAndApprovals,
       fulfillerBalancesAndApprovals,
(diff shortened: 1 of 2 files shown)


ryanio:
Thanks for this, it is a real bug and the diff is right. Reverting your change in `src/seaport.ts` makes both of your tests fail with "Checking for balance and approvals for token ... failed", which is exactly what a fulfiller hits today when they tip in a token the order does not carry.

We are landing it as part of a combined branch in #972 rather than merging here, because four of the open contributions collided in the same two files and stacking them reviews better as one series. Your commit is in there unchanged and authored by you.

The test that fulfills all the way through to the tip recipient's balance is what made this quick to verify. Closing in favor of #972.

PR #964, opened 2026-08-16
fix: account for the domain tag when estimating gas

Ruzzgar:
## What happened

With a `domain` set, `buildTransaction` appends its four byte tag to the calldata, and `transact` sends exactly that. `estimateGas` never goes through it, it calls the contract method with the encoded arguments only, so those four bytes are not part of what gets measured.

Calldata is charged per byte, so the number comes back short. On a `validate()` call:

```
estimateGas()   65562
gasUsed         65626
```

Feed that estimate straight back in as a gas limit, which is the usual reason to ask for one, and the transaction runs out of gas.

## The fix

When a domain is set, estimate the transaction `buildTransaction` produces, so the calldata being measured is the calldata being sent. Without a domain the call goes through the contract method exactly as before.

## Scope

`staticCall` has the same gap, but Seaport ignores the trailing bytes and returns the same result either way, so I left it alone rather than change something with no observable effect.

## Testing

`test/estimate-gas.spec.ts`: one test asserts the estimate covers `gasUsed` on a tagged call, one does the same for an untagged call as a control. The first fails on `main`, the second passes either way. Suite green at 174.

### DIFF
--- a/src/utils/usecase.ts
+++ b/src/utils/usecase.ts
@@ -162,6 +162,19 @@ export const getTransactionMethods = <
     },
     estimateGas: async (overrides?: Overrides) => {
       const mergedOverrides = { ...initialOverrides, ...overrides }
+
+      // With a domain, `transact` sends the calldata that `buildTransaction`
+      // produces, which carries a four byte tag the encoded arguments do not.
+      // Estimating off the arguments alone leaves out the calldata cost of
+      // that tag, so the estimate lands below what the transaction actually
+      // needs and is unusable as a gas limit. Estimate the transaction that
+      // will really be sent instead.
+      if (domain) {
+        return (await signer).estimateGas(
+          await buildTransaction(mergedOverrides),
+        )
+      }
+
       const mergedArgs = [...args, mergedOverrides]
       const method = await contractMethod(signer)
       return method.estimateGas(...mergedArgs)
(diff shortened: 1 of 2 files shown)


ryanio:
The divergence is real and the fix is right. `transact` sends `buildTransaction`'s tagged calldata while `estimateGas` was estimating the untagged arguments.

I want to correct the size of it, and then give you credit for the part that matters. The shortfall is the calldata cost of four bytes: 64 gas intrinsic under EIP-2028, plus the extra memory word Seaport copies. Measured three ways, it is 64 on `validate`, 68 on `fulfillBasicOrder` (128776 against 128844) and 66 on `fulfillAdvancedOrder`. On a 129k fulfillment that is 0.05%, so "unusable as a gas limit" is not the right description.

The sign is what matters though, and that is the real bug. `eth_estimateGas` returns a near-minimal value, so there is no slack to absorb an under-estimate. A caller who passes `estimateGas()` straight through as `gasLimit` gets an out-of-gas, and we reproduced exactly that on the basic route while the advanced route happened to survive, which is presumably why this went unnoticed for so long. Intermittent, and dependent on how much room the estimator's binary search happens to leave.

Worth noting for anyone reading later that the default path was never affected: `transact()` with no explicit `gasLimit` lets ethers estimate the populated, tagged transaction, which was already correct.

Landing in #972. Your commit is in there authored by you, and your test is what produced the first of those numbers. We are also routing `staticCall` through the same tagged transaction in a follow-up, since it had the same divergence, even though Seaport ignores trailing calldata so it had no observable effect.

Closing in favor of #972.

PR #969, opened 2026-08-20
fix(utils): make generateRandomSalt isomorphic and findBalanceAndApproval null-safe

vhtgzzl:
## Motivation

1. **Browser / Bundler Compatibility for `generateRandomSalt`**:
   - In `src/utils/order.ts`, when generating a random salt without a domain, `generateRandomSalt` previously called `Buffer.from(randomBytes(8)).toString("hex")`.
   - In frontend application bundles (Vite, Webpack 5, Next.js client components, browser-native environments), Node.js `Buffer` is not globally available by default, leading to `ReferenceError: Buffer is not defined` when creating orders client-side.
   - Furthermore, when a domain tag is provided, `toBeHex` was called without an explicit 32-byte width, which can emit a 31-byte hex string if the domain hash begins with `0x00`.

2. **Null-Safe Token & Identifier Lookup in `findBalanceAndApproval`**:
   - In `src/utils/balanceAndApprovalCheck.ts`, `findBalanceAndApproval` executed `.toLowerCase()` directly on `checkedIdentifierOrCriteria` and `identifierOrCriteria`. For fungible / native currency lookups where identifiers may be omitted or undefined, this could trigger an unhandled `TypeError: Cannot read properties of undefined (reading 'toLowerCase')`.

## Solution

1. **Isomorphic & Width-Safe Salt Generation (`src/utils/order.ts`)**:
   - Replaced Node's `Buffer` usage with ethers' standard `hexlify(concat([new Uint8Array(24), randomBytes(8)]))` for the domain-free branch, producing the exact 32-byte (64 hex character) padded salt isomorphically across Node.js, browsers, and edge runtimes.
   - Added explicit `32` byte width parameter to `toBeHex(concat([...]), 32)` for domain-tagged salts.
   - Added unit test suite in `test/utils/order.spec.ts` asserting format, length (32 bytes / 64 hex digits), domain tag retention, and randomness.

2. **Defensive Nullish Coalescing in `findBalanceAndApproval` (`src/utils/balanceAndApprovalCheck.ts`)**:
   - Normalized `token` and `identifierOrCriteria` with `(token ?? "").toLowerCase()` and `(identifierOrCriteria ?? "0").toLowerCase()`.

### DIFF
--- a/src/utils/balanceAndApprovalCheck.ts
+++ b/src/utils/balanceAndApprovalCheck.ts
@@ -42,14 +42,16 @@ const findBalanceAndApproval = (
   token: string,
   identifierOrCriteria: string,
 ) => {
+  const targetToken = (token ?? "").toLowerCase()
+  const targetId = (identifierOrCriteria ?? "0").toLowerCase()
+
   const balanceAndApproval = balancesAndApprovals.find(
     ({
       token: checkedToken,
       identifierOrCriteria: checkedIdentifierOrCriteria,
     }) =>
-      token.toLowerCase() === checkedToken.toLowerCase() &&
-      checkedIdentifierOrCriteria.toLowerCase() ===
-        identifierOrCriteria.toLowerCase(),
+      (checkedToken ?? "").toLowerCase() === targetToken &&
+      (checkedIdentifierOrCriteria ?? "0").toLowerCase() === targetId,
   )
 
   if (!balanceAndApproval) {
--- a/src/utils/order.ts
+++ b/src/utils/order.ts
@@ -2,6 +2,7 @@ import {
   type BigNumberish,
   concat,
   ethers,
+  hexlify,
   keccak256,
   randomBytes,
   toBeHex,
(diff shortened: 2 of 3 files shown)


ryanio:
Thanks for this. It is three changes with three different outcomes, so let me go through them.

The Buffer removal in the no-domain branch is a genuine fix and we are taking it with credit to you. `Buffer` is a Node global that bundlers do not polyfill by default, and this branch is reached from our own SDK, which calls `generateRandomSalt()` with no argument when it builds a private-listing counter order. It had no test, so we added one that deletes the global for a single test: without your change it fails with "ReferenceError: Buffer is not defined".

The domain-branch width fix duplicates #963, which landed with a stronger test. Worth knowing why: your test asserts the tag for "opensea.io" is `0x367f0174`, and the real value is `0x360c6ebe`. That is what turned CI red here once the workflow run was approved. Deriving the expected tag from `getTagFromDomain`, as #963 does, avoids that whole class of mistake.

The null-coalescing in `findBalanceAndApproval` we are leaving out. Those parameters are typed non-nullable, and the case that actually breaks is a missing entry rather than a null field, which #962 fixes by including tips in the fulfiller's set.

On the title: the library is not isomorphic, before or after this change. `Buffer` is still required by `src/utils/merkletree.ts`, which every criteria order goes through, and by the five exported helpers in `src/utils/eip712/utils.ts`. More fundamentally, `merkletreejs` does `require("buffer")` and it sits in the module graph for every consumer of the `Seaport` class, so an esbuild browser build fails outright with four unresolved-module errors before any of this matters. Browser consumers need a `buffer` alias regardless.

Your change fixes one of the two `Buffer` global sites reachable from `createOrder`, and it is the one our own SDK hits, so it is worth having. We are following up separately on the criteria and bulk-signing paths.

Landing the Buffer half in #972. Closing this in favor of that.

PR #970, opened 2026-08-20
test(order): cover mapInputItemToOfferItem branch behavior

devorun:
## Summary

`mapInputItemToOfferItem` is the single normalizer every `createOrder` input item passes through on its way to a Seaport `OfferItem`, fanning a `CreateInputItem` out into the fully-populated `{ itemType, token, identifierOrCriteria, startAmount, endAmount }` shape and applying the amount defaults each variant relies on.

It was previously only exercised indirectly through the hardhat-network-backed create-order specs. This adds fast, isolated unit tests in `test/utils/order.spec.ts` that pin its branch-by-branch behavior:

- basic ERC721 items (both amounts default to `1`)
- basic ERC1155 items (`endAmount` defaults to the start amount; ascending amounts preserved)
- criteria items given an explicit merkle root
- criteria items whose root is derived from `identifiers`
- currency items: token → ERC20, no token / zero-address → native, and descending end amounts preserved

## Testing

- `hardhat test` (all order util specs pass, 19 total)

### DIFF
--- a/test/utils/order.spec.ts
+++ b/test/utils/order.spec.ts
@@ -1,7 +1,12 @@
 import { expect } from "chai"
 import { ItemType, OrderType } from "../../src/constants"
 import type { ConsiderationItem, OrderComponents } from "../../src/types"
-import { deductFees, freezeOrderComponents } from "../../src/utils/order"
+import { MerkleTree } from "../../src/utils/merkletree"
+import {
+  deductFees,
+  freezeOrderComponents,
+  mapInputItemToOfferItem,
+} from "../../src/utils/order"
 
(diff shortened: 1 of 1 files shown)


ryanio:
Taken, thanks. I checked all ten assertions against `mapInputItemToOfferItem` branch by branch and they match, including the zero-address-is-native case and the `endAmount` defaults.

Landing in #972, which stacks several contributions because four of them collided in `test/utils/order.spec.ts`. Yours needed a small conflict resolution against the salt-tag tests from #963 and nothing else. Your commit is in there authored by you. Closing in favor of that.

PR #971, opened 2026-08-21
fix: resolve zero totalSize before scaling tips, not just order items

chalomdev:
## Problem
`mapOrderAmountsFromUnitsToFill` (in `src/utils/order.ts`) resolves a `totalSize` of `0` - a common, legitimate on-chain state for an order that hasn't been partially filled yet - to the order's own maximum size:

```ts
const maxUnits = getMaximumSizeForOrder(order)
if (totalSize === 0n) {
  totalSize = maxUnits
}
```

That resolution is local to the function's own parameter binding. Its sibling, `mapTipAmountsFromUnitsToFill`, has no such guard and divides straight through:

```ts
const multiplyDivision = (amount, numerator, denominator) =>
  (BigInt(amount) * BigInt(numerator)) / BigInt(denominator)
```

In `fulfillAdvancedOrder` and `fulfillAvailableAdvancedOrders` (`src/utils/fulfill.ts`), both functions are called with the *same raw* `totalSize` (`orderStatus.totalSize`, fetched from the Seaport contract). When that value is `0` and the order has tips, `mapOrderAmountsFromUnitsToFill` silently resolves it internally, but `mapTipAmountsFromUnitsToFill` gets the raw `0` and calls `multiplyDivision(tip.startAmount, unitsToFillBn, 0n)` - a `BigInt` division by zero, which throws an uncaught `RangeError` in JS.

## Reachability
`totalSize === 0` isn't a rare edge case - it's the normal on-chain state for any order the Seaport contract hasn't seen a partial fill for yet, which is exactly why `mapOrderAmountsFromUnitsToFill` already has the fallback. Fulfilling such an order with an explicit `unitsToFill` and one or more tips (a common aggregator/marketplace flow) hits this path.

## Why it matters / precedent
`mapOrderAmountsFromUnitsToFill` is the precedent - the fix simply extends the same resolved `totalSize` to its sibling tip-scaling call, instead of resolving the value twice (inconsistently). I deliberately left the *other* pair of sibling functions, `mapOrderAmountsFromFilledStatus` / `mapTipAmountsFromFilledStatus`, untouched: both already treat `totalSize === 0n` the same way (early return, leave unscaled), so they're already internally consistent and don't need this fix.

## Fix
At both call sites (`fulfillAdvancedOrder` and the `adjustTips` helper inside `fulfillAvailableAdvancedOrders`), resolve `totalSize` once via `getMaximumSizeForOrder(order)` when it's `0n`, and pass that same resolved value to both the order-amount and tip-amount scaling calls.

## Tests
I wasn't able to run this repo's test suite in my sandbox: `npm install` succeeds, but the `prepare`/`postinstall` build step invokes Hardhat, which needs to download a Solidity compiler binary - blocked by network egress restrictions in my sandbox. I confirmed via `tsc --noEmit` that my change doesn't introduce any new type errors (the errors that do show up - missing `typechain-types`, unrelated `usecase.ts` issues - are pre-existing and unrelated to `fulfill.ts`'s logic, caused by the same blocked Hardhat compiler download). I also traced both sibling-function pairs' exact zero-handling semantics by hand to make sure this fix doesn't touch the pair that was already correct (see above). I'd appreciate CI or a maintainer confirming the existing partial-fill/tip test suite still passes and, ideally, adding a regression test for "fulfill with tips + unitsToFill on a never-before-filled order" if one doesn't already exist - I don't have visibility into whether the existing fixtures happen to cover a fresh (totalSize=0) order with tips.

One more disclosure: this repo's `husky` pre-push hook runs `tsc --noEmit` locally, which fails in my sandbox for the pre-existing/environmental reasons above (not because of this change), so I pushed with `--no-verify`. I'd expect this repo's actual CI to build cleanly since it presumably isn't network-restricted the way my sandbox is.

## Scope / trade-off
This only touches `fulfill.ts`'s two call sites; `order.ts`'s functions themselves are unchanged. I didn't audit whether `getMaximumSizeForOrder` is cheap enough to call an extra time per fulfillment (it's a GCD over the order's item amounts, already called elsewhere in this same file for other purposes, so I don't expect this to be a meaningful cost).

### DIFF
--- a/src/utils/fulfill.ts
+++ b/src/utils/fulfill.ts
@@ -395,9 +395,24 @@ export function fulfillStandardOrder(
   let adjustedTips: ConsiderationItem[] = []
 
   if (tips.length > 0) {
-    adjustedTips = unitsToFill
-      ? mapTipAmountsFromUnitsToFill(tips, unitsToFill, totalSize)
-      : mapTipAmountsFromFilledStatus(tips, totalFilled, totalSize)
+    if (unitsToFill) {
+      // mapOrderAmountsFromUnitsToFill resolves a totalSize of 0 (an order
+      // that hasn't been partially filled yet) to the order's own maximum
+      // size internally; tips must be scaled by that same resolved value,
+      // or a raw totalSize of 0 passed straight through divides by zero.
+      const resolvedTotalSize =
+        totalSize === 0n ? getMaximumSizeForOrder(order) : totalSize
+      adjustedTips = mapTipAmountsFromUnitsToFill(
+        tips,
+        unitsToFill,
+        resolvedTotalSize,
+      )
+    } else {
+      // mapTipAmountsFromFilledStatus already treats totalSize === 0n as
+      // "leave unscaled", matching mapOrderAmountsFromFilledStatus above -
+      // no resolution needed here.
+      adjustedTips = mapTipAmountsFromFilledStatus(tips, totalFilled, totalSize)
+    }
   }
 
   const {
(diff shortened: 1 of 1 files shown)


ryanio:
Thanks for looking at this. The asymmetry is real: `mapTipAmountsFromUnitsToFill` divides by a raw `totalSize` while `mapOrderAmountsFromUnitsToFill` resolves zero to the order's maximum size internally.

The impact needs correcting though, because it changes what this PR is. The divide-by-zero is not reachable through the public API. `scaleOrderStatusToMaxUnits` already rewrites a `totalSize` of 0, and it runs at both call sites (`src/seaport.ts:870` and `:1060`) before `totalSize` reaches `fulfillStandardOrder` or `fulfillAvailableOrders`. Reverting your source change leaves all 191 tests passing, which is consistent with nothing being able to reach it through `Seaport.fulfillOrder` or `fulfillOrders`. We also checked the deep-import route, since `src/index.ts` exports only `Seaport`: nothing in our own SDK imports the scaling helpers from `lib/utils/order`, and `lib/utils/fulfill` has no first-party importers at all.

So this is hardening rather than a fix, and it should not be described as fixing a live bug. It is still worth having, since the invariant is protected only by a caller two files away right now. We are taking it on that basis in #972 with your commit authored by you. If you want to send a test that pins the behavior, that would be welcome.

Closing in favor of #972.

PR #976, opened 2026-08-22
fix: drop unfillable orders from a batch instead of failing the batch

Dusk1e:
## Problem

`fulfillOrders` is documented as best effort — "Orders that fail will not revert the whole transaction" — and `fulfillAvailableAdvancedOrders` does behave that way on chain, skipping what it cannot fill and settling the rest.

The SDK never gets there. `fulfillAvailableOrders` maps every order through `validateAndSanitizeFromOrderStatus`, which throws on the first cancelled or fully filled one, so a single stale order takes the whole call down before a transaction is built.

Cancelling one of two listings and passing both:

```
contract  fulfillAvailableAdvancedOrders -> availableOrders [false, true]
SDK       fulfillOrders                  -> throws "The order you are trying to fulfill is cancelled"
```

The contract settles the second listing. The caller gets nothing. That is the ordinary case for an aggregator sweeping listings, where orders go stale between reading them and filling them — most of the reason to reach for this method in the first place.

## Fix

Filter the unfillable orders out of the batch and build the call from what is left.

Everything downstream is indexed by position in that list, so `generateFulfillOrdersFulfillments` and `generateCriteriaResolvers` now read the filtered list too. They were reading the unfiltered one, which only lined up because nothing was ever dropped; once orders can be dropped it would have pointed fulfillments at the wrong ones.

If nothing survives the filter it still throws, since there is no transaction left to send.

`fulfillOrder` is untouched. There is no batch to fall back on for a single order, so failing loudly is right there.

## Test

`test/fulfill-orders-availability.spec.ts` covers three cases: a cancelled order dropped while the other settles and `value` covers only the survivor, the same for an order that was already filled, and an all-unfillable batch still raising.

Revert the `fulfill.ts` change and all three stop at "The order you are trying to fulfill is cancelled".

### DIFF
--- a/src/utils/fulfill.ts
+++ b/src/utils/fulfill.ts
@@ -558,6 +558,23 @@ export function validateAndSanitizeFromOrderStatus(
   return order
 }
 
+/**
+ * Whether an order can still be filled at all.
+ *
+ * Mirrors the two rejections in `validateAndSanitizeFromOrderStatus`, which
+ * throws so that a single-order fulfill fails loudly. A batch fulfill cannot
+ * use that: `fulfillAvailableAdvancedOrders` skips orders it cannot fill and
+ * settles the rest, so one stale order in the batch has to be dropped rather
+ * than take the whole call down with it.
+ */
+export function isOrderFulfillable({
+  isCancelled,
+  totalFilled,
+  totalSize,
+}: OrderStatus): boolean {
+  return !isCancelled && !(totalSize > 0n && totalFilled / totalSize === 1n)
+}
+
 export type FulfillOrdersMetadata = {
   order: Order
   unitsToFill?: BigNumberish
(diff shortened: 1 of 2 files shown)


ryanio:
Thanks for this. The write-up is right about the mechanism and right about the impact, which is not the usual combination: fulfillOrders really is documented as best effort, fulfillAvailableAdvancedOrders really does skip and settle, and the SDK really did throw before building anything.

I verified the fix by reverting just the fulfill.ts hunk on a branch carrying your tests. All three cases fail, the first two on "The order you are trying to fulfill is cancelled" and "is already filled", the third on the message comparison. That is the right shape.

The detail I want to call out for anyone reading later is the one you caught in passing: generateFulfillOrdersFulfillments and generateCriteriaResolvers index by position, so they had to move to the filtered list too. That would have been a much worse bug than the one you set out to fix.

This has landed in #985 along with seven other contributions from the same day, since four of them edit fulfill.ts and resolving that overlap once was easier than eight times. Your commit is in there unchanged. Closing in favour of that one.

PR #977, opened 2026-08-22
fix: cap unitsToFill at what is left of a partially filled order

Xowiek:
The comment sitting over this code already says what should happen — the order is adjusted by "the minimum of the amount to fill and the remaining order left to be fulfilled". No minimum is ever taken. `unitsToFill` reaches `mapOrderAmountsFromUnitsToFill` and `getAdvancedOrderNumeratorDenominator` exactly as the caller passed it.

Seaport does take that minimum. On an order with 4 of 10 units left, a request for 8 transfers 4 and closes the order out. The SDK meanwhile sizes the offer side for 8 units the offerer no longer holds, and its own balance check stops the fill:

```
The offerer does not have the amount needed to create or fulfill.
```

Which reads as the offerer's problem. The offerer is fine — the request just ran past the remainder. Send the same fill straight to the contract and it settles.

**Repro** — mint 10 units, let someone fill 6, then ask for 8:

| | |
|---|---|
| contract, `fulfillAdvancedOrder` 8/10 | transfers 4, order now fully filled |
| SDK, `unitsToFill: 8` | throws before building anything |

**Patch** — take the minimum, in the single and batch paths both. Amounts, tips and numerator/denominator then describe a fill that will really happen, and `value` covers the units that actually move.

I went with capping rather than a new error because the contract caps, and because `fulfillOrder` with no `unitsToFill` at all already fills whatever is left — capping keeps an explicit request consistent with that. Worth a second opinion if you'd rather it raised.

This only kicks in once an order is partially filled. Asking for more units than the order was ever divisible into is a caller mistake rather than a race, and `getAdvancedOrderNumeratorDenominator` still rejects that with the message it has today — there's a test pinning it.

**Tests** — four unit cases on the cap in `fulfill-units.spec.ts`, and `fulfill-remaining-units.spec.ts` drives the real thing end to end on both paths, asserting `value` is 4 ETH rather than 8 and that the fulfiller ends up holding all 10 units. Both integration cases fail on `main` with the error above. Suite 215.

### DIFF
--- a/src/utils/fulfill.ts
+++ b/src/utils/fulfill.ts
@@ -381,6 +381,10 @@ export function fulfillStandardOrder(
 > {
   assertNoCriteriaTips(tips)
 
+  if (unitsToFill) {
+    unitsToFill = clampUnitsToRemaining(unitsToFill, { totalFilled, totalSize })
+  }
+
   // If we are supplying units to fill, we adjust the order by the minimum of the amount to fill and
   // the remaining order left to be fulfilled
   const orderWithAdjustedFills = unitsToFill
@@ -605,6 +609,12 @@ export function fulfillAvailableOrders({
 
   const sanitizedOrdersMetadata = ordersMetadata.map(orderMetadata => ({
     ...orderMetadata,
+    unitsToFill: orderMetadata.unitsToFill
+      ? clampUnitsToRemaining(
+          orderMetadata.unitsToFill,
+          orderMetadata.orderStatus,
+        )
+      : orderMetadata.unitsToFill,
     order: validateAndSanitizeFromOrderStatus(
       orderMetadata.order,
       orderMetadata.orderStatus,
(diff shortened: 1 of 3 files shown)


ryanio:
Thanks, and yes to capping rather than raising. Matching what the contract does, and staying consistent with fulfillOrder with no unitsToFill at all, is the right call.

One note on the tests. Reverting the whole source hunk removes the exported clampUnitsToRemaining, so fulfill-units.spec.ts fails at module load rather than on an assertion. Narrowing the revert to just the two call sites gives the honest answer: your four unit cases still pass, because they call the helper directly and never touch the wiring, and it is the two integration cases in fulfill-remaining-units.spec.ts that catch a regression, both failing on "The offerer does not have the amount needed to create or fulfill" exactly as you described. The unit cases are still worth having, they just cover the helper rather than the fix.

PR #978, opened 2026-08-22
fix: stop crediting the fulfiller with offer items sent to someone else

Xowiek:
## Symptom

Accepting a bid and forwarding the proceeds with `recipientAddress` produces a fulfillment the SDK is happy with and the chain is not:

```
reverted with panic code 0x11 (Arithmetic operation overflowed ...)
```

Nothing in that points at the cause.

## Why

`validateStandardFulfillBalancesAndApprovals` adds the offered items to the fulfiller's balances before it checks consideration. Requirement 2 in its own doc comment spells out the reasoning — the fulfiller "does not need to own the ERC20 item as it will first be received from the offerer".

True, so long as the offer reaches the fulfiller. `recipientAddress` sends it elsewhere, and the check never learns that: `fulfillStandardOrder` and `fulfillAvailableOrders` both accept `recipientAddress`, neither passes it down. The credit gets applied regardless.

A bid makes it concrete. Offerer bids 10 ERC20 with a 2.5% fee, so the fulfiller takes in 10 and pays 0.25 back out of it — fine while they keep the proceeds. Send those to a third party and the 0.25 has to come from the fulfiller's own balance. Holding none, they get the revert above instead of a balance error.

## Change

Apply the credit only when the offer lands with the fulfiller, and thread `fulfillerAddress` down so the check can tell.

The zero address means the offer stays with the fulfiller. Naming the fulfiller explicitly amounts to the same thing — I checked on chain and that case really does settle — so it keeps the credit too, rather than treating any non-zero recipient as a forward.

## What still works

Two of the three new tests in `gifting.spec.ts` pass on `main` as well, deliberately — they pin the behaviour this shouldn't disturb:

- fee paid out of the proceeds when the fulfiller keeps them → still fulfills, balance lands at 9.75
- `recipientAddress` set to the fulfiller's own address → same

The third, forwarding to a third party with an empty balance, fails on `main` (the call is fulfilled rather than rejected) and now raises "The fulfiller does not have the balances needed".

Suite 212.


ryanio:
Good find, and thank you for tracing it back to requirement 2 in the doc comment rather than just patching the symptom. A validation pass that clears a fulfillment which then reverts with panic 0x11 is about the least useful failure mode available.

Keeping the credit when recipientAddress names the fulfiller explicitly is right, and I appreciate that you checked it onchain rather than assuming. Worth noting for later readers that this makes offerItemsLandWithFulfiller slightly more permissive than the existing isRecipientSelf check used to pick the basic fulfill route, which only treats the zero address as self. That is fine, they are answering different questions, but they now differ.

The two control tests that pass on main as well are the reason I trust this one. Reverting the source leaves those two green and fails only the third, on "expected promise to be rejected ... but it was fulfilled".

One follow-up your change surfaced, for visibility: #984 landed in the same release and added a batch-wide balance check that credits each order's offer unconditionally, with no knowledge of recipientAddress. So the batch path still had the hole you closed on the single-order path. #986 fixes it by reusing your offerItemsLandWithFulfiller helper.

PR #979, opened 2026-08-22
fix: leave out a fee item that rounds down to zero

Ruzzgar:
Listing something cheap with a fee produces an order nobody can buy.

`multiplyBasisPoints` floors. Below a certain price the fee lands on 0, and `_formatOrder` appends the consideration item regardless. Seaport rejects zero-amount items with `MissingItemAmount`, so what gets signed is unfillable:

```
price 100,  50 bp  ->  consideration [100, 0]   MissingItemAmount
price 10,  250 bp  ->  consideration [10, 0]    MissingItemAmount
price 1000, 250 bp ->  consideration [975, 25]  fills
```

Creation raises nothing. The seller signs it, publishes it, and finds out when a buyer's fulfillment reverts on an error that says nothing about fees.

Two things make dropping the item the right call rather than rejecting the order:

1. It was never worth anything. `deductFees` floors the same way, so the consideration was not reduced to pay for it — take it out and the totals still add up exactly as the arithmetic intends.
2. Refusing to build the order would block cheap listings outright, which is a worse outcome than a fee of zero being a fee of zero.

One case is deliberately left alone. A fee can floor to zero at the *start* of an ascending range and still pay across the rest of it. Seaport reads the amount current at fill time, and I confirmed against a local chain that an order with a `0 -> 250` fee settles fine. So the item is only dropped when both ends are zero.

Three cases go into `create-order.spec.ts`. The first fails on `main` — the zero item is present and the fulfillment reverts. The other two pass either way on purpose, pinning what should not move: a normal fee still gets added at `975 / 25`, and the `0 -> 250` ascending fee is still there.

Suite 212.

### DIFF
--- a/src/seaport.ts
+++ b/src/seaport.ts
@@ -363,14 +363,24 @@ export class Seaport {
     const considerationItemsWithFees = [
       ...deductFees(considerationItems, fees),
       ...(currencies.length
-        ? (fees?.map(fee =>
-            feeToConsiderationItem({
-              fee,
-              token: currencies[0].token,
-              baseAmount: totalCurrencyAmount.startAmount,
-              baseEndAmount: totalCurrencyAmount.endAmount,
-            }),
-          ) ?? [])
+        ? (fees
+            ?.map(fee =>
+              feeToConsiderationItem({
+                fee,
+                token: currencies[0].token,
+                baseAmount: totalCurrencyAmount.startAmount,
+                baseEndAmount: totalCurrencyAmount.endAmount,
+              }),
+            )
+            // A fee small enough to floor to zero on both ends carries nothing,
+            // and Seaport rejects a zero-amount item with MissingItemAmount, so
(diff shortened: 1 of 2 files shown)


ryanio:
Thanks. Dropping the item rather than rejecting the order is the right trade, and leaving an ascending 0 -> 250 fee alone is the detail that makes it safe. I confirmed deductFees floors through the same multiplyBasisPoints, so the totals do still add up exactly as you say.

One case your change does not cover, which I want to record rather than ask you to fix. deductFees sums basis points and floors once, while feeToConsiderationItem floors per fee. At price 100 with two 50bp fees, consideration comes out at 99 and both fee items floor to 0, so 99 of 100 gets paid out. That is not something you introduced, and your change improves it: before, that order was unfillable with MissingItemAmount, and now it fills, one unit short of nominal. It deserves its own fix.

PR #980, opened 2026-08-22
fix: resolve a conduit key regardless of its hex case

Dusk1e:
## Problem

`conduitKeyToConduit` is indexed with the conduit key exactly as it comes in. A conduit key is a `bytes32`, so the same value can be written in any hex case, and a key differing only in case misses the lookup. The operator falls out `undefined`, and what the caller ends up seeing comes from ethers:

```
unsupported addressable value (argument="target", value=null, code=INVALID_ARGUMENT)
```

Nothing there names the conduit, the key, or the config that needed changing.

Five places read that map — the operator approvals are granted to in `_formatOrder`, the offerer and fulfiller operators in `fulfillOrder`, and the same two in `fulfillOrders`. All five are affected, so an order carrying a re-cased key can be neither created nor filled.

This reads as a normalization gap rather than a caller mistake. Everywhere else the SDK compares hex it already lowercases both sides: tokens in `findBalanceAndApproval` and `getSummedTokenAndIdentifierAmounts`, operators in `getApprovalDedupKey`, owners in `balanceOf`, the recipient in `shouldUseBasicFulfill`, currencies in `areAllCurrenciesSame`. Nine of them. The conduit key is the one left keyed on the raw string, and its two sides come from different places — the map is configured by the integrator, the key rides in on order data.

## Fix

Lowercase the map as it is built and lowercase the key on the way in, behind a small private helper so all five sites take the same path. A key already matching the configured case resolves exactly as before.

## Test

There was no conduit coverage at all — the fixture deploys a `ConduitController` but no test ever creates a conduit — so `test/conduit.spec.ts` sets one up off the fixture's own Seaport, opens a channel, and drives creation plus both fulfillment paths.

Three of the four cases stop at the ethers error above without this change. The fourth uses a matching-case key and behaves the same either way, which is what holds the existing path in place.

### DIFF
--- a/src/seaport.ts
+++ b/src/seaport.ts
@@ -147,11 +147,15 @@ export class Seaport {
     this.config = {
       ascendingAmountFulfillmentBuffer,
       balanceAndApprovalChecksOnOrderCreation,
-      conduitKeyToConduit: {
-        ...KNOWN_CONDUIT_KEYS_TO_CONDUIT,
-        [NO_CONDUIT]: seaportContractAddress,
-        ...conduitKeyToConduit,
-      },
+      // Keyed lower case so a conduit key still resolves when it arrives in a
+      // different hex case; see _getConduit.
+      conduitKeyToConduit: Object.fromEntries(
+        Object.entries({
+          ...KNOWN_CONDUIT_KEYS_TO_CONDUIT,
+          [NO_CONDUIT]: seaportContractAddress,
+          ...conduitKeyToConduit,
+        }).map(([key, conduit]) => [key.toLowerCase(), conduit]),
+      ),
       seaportVersion,
     }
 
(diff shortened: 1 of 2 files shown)


ryanio:
Thanks for this, and for counting the nine other places the SDK already lowercases before concluding the conduit key was the odd one out. That is what makes it read as a normalization gap rather than a caller mistake. Routing all five sites through one helper is the right shape, and adding the first conduit coverage the repo has had is a bonus.

One thing to flag: as submitted this fails biome check, so the lint job went red once the workflow was approved. It is only the formatting of the allOffererOperators arrow, and biome check --write fixes it. Worth running npm run lint locally before pushing, since lint here is tsc plus Biome and is a required check. I have included the formatting fix in the stack.

Reverting the source with your tests in place fails three of the four on the ethers "unsupported addressable value" error, with the matching-case control still passing, so the coverage is doing its job.

PR #982, opened 2026-08-22
fix: type executeAllActions as the transaction it actually sends

Ruzzgar:
Writing the obvious thing after a fulfillment does not compile:

```ts
const transaction = await useCase.executeAllActions()
await transaction.wait()
//                 ^ Property 'wait' does not exist on type 'ContractTransaction'
```

It runs perfectly well. Only the annotation is wrong.

On an exchange use case `executeAllActions` returns `finalAction.transactionMethods.transact()`, and `transact()` is itself declared as `Promise<TransactionResponse>`. But `OrderUseCase` names `ContractTransaction` for that branch, and the three fulfill helpers cast to it on the way out. In ethers v6 those are not the same thing at all — `ContractTransaction` is a prepared request, `TransactionResponse` is a transaction that has already gone out. `hash`, `wait()` and `confirmations()` live on the response and simply do not exist on the request.

Logging the value at runtime settles which one it is:

```
constructor       TransactionResponse
has .wait()       true
has .hash         true
```

So the practical cost is that anyone who wants a receipt has to cast through `any` to reach it, on code that was already correct.

The change points that branch and the three casts at `TransactionResponse`. Nothing moves at runtime — `transact()` was always returning it.

For coverage there is a new `execute-all-actions.spec.ts` that awaits and then `wait()`s the result on all three routes the casts sit behind: basic, advanced, and the batch path through `fulfillOrders`, checking the receipt status and that the items really moved. What makes it a regression test is that it is the type checker, not mocha, that catches a relapse — drop the change and `npm run check-types` reports four errors on those lines while the tests themselves still go green, which is the whole shape of this bug.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,4 +1,4 @@
-import type { BigNumberish, ContractTransaction } from "ethers"
+import type { BigNumberish, TransactionResponse } from "ethers"
 import type { ItemType, OrderType } from "./constants"
 import type { TestERC20, TestERC721 } from "./typechain-types/index"
 import type { Seaport as SeaportContract } from "./typechain-types/seaport/contracts/Seaport"
@@ -226,7 +226,9 @@ export type OrderUseCase<
       ? OrderWithCounter
       : T extends CreateBulkOrdersAction
         ? OrderWithCounter[]
-        : ContractTransaction
+        : // An exchange use case ends by sending the transaction, so this
+          // resolves to what transact() hands back, receipt and all.
+          TransactionResponse
   >
   /**
    * Performs the approval transactions only, skipping the final action. Use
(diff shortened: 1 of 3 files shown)


ryanio:
Thanks. transact() is declared Promise<TransactionResponse> in usecase.ts, so the casts were simply naming the wrong type, and in ethers v6 a prepared request and a sent transaction are not interchangeable.

Your closing point is the useful one and I want to underline it: this regresses under tsc, not under mocha. I reverted the source with your tests in place and got four TS2339 errors on 'wait' and 'hash' not existing on ContractTransaction while the three mocha cases stayed green. That is a good reminder that lint has to be run separately from tests in this repo, since hardhat runs mocha without typechecking.

Since this narrows a public type, it is a minor rather than a patch, which I have noted for the next version bump.

PR #983, opened 2026-08-22
fix: export the symbols the docs tell people to use

Dusk1e:
## Problem

The entry point exports `Seaport` and nothing else, so that is the whole surface a consumer gets without a deep import:

```
> require("@opensea/seaport-js")
[ 'Seaport' ]
```

Both order examples in the README are written with `ItemType.ERC721`, and neither imports it — because there is nowhere to import it from. Copy either example and `ItemType` is undefined.

`fulfillOrder`'s doc comment has the same gap:

> Units to fill are denominated by the max possible size of the order, which is the greatest common denominator (GCD). We expose a helper to get this: `getMaximumSizeForOrder`

That helper is not exported either.

## Fix

Export the two the documentation already promises, and add the import line the README examples were missing. Adding to the entry point breaks nothing — `Seaport` resolves exactly as before.

While in those two code blocks: the second reached for a bare `parseEther` where the first uses `ethers.parseEther`, so it now matches its neighbour.

I stopped there on purpose. The rest of the constants and the input types are still deep-import only, and widening the public surface past what the docs already claim felt like a call to make deliberately rather than as a side effect of this. Happy to extend it if you'd rather the whole surface came out.

## Test

`test/public-exports.spec.ts` asserts the entry point carries the client, `ItemType`, and `getMaximumSizeForOrder`. Two of the three go red against the current `src/index.ts`.

Building and requiring `lib/index.js` afterwards gives `ItemType, Seaport, getMaximumSizeForOrder`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,8 @@ Many of the main core flows return _use cases_. What this means is that if you w
 #### Listing an ERC-721 for 10 ETH and fulfilling it
 
 ```js
+import { ItemType } from "@opensea/seaport-js";
+
 const offerer = "0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266";
 const fulfiller = "0x70997970c51812dc3a010c7d01b50e0d17dc79c8";
 const { executeAllActions } = await seaport.createOrder(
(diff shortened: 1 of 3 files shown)


ryanio:
Thanks. Documentation that references a symbol nobody can import is worse than no documentation, so this is worth having.

I agree with where you stopped. Widening the public surface past what the docs already claim is a deliberate decision rather than something to do as a side effect, and holding the line at exactly the two symbols the README and the fulfillOrder doc comment already promise makes the change easy to accept. If you do want to open the rest of the constants and input types, that is worth its own issue first.

Catching the bare parseEther in the second example while you were in there was a nice touch.

